PR #113, opened 2022-05-24
feat(iam): add structured logging to IAM package

dhx12:
This adds an intitial pass as structured logging using Winston to help with debugging in deployed
environments

Only added logging to providers for now, but logger abstraction in utils could be
used elsewhere.

### DIFF
--- a/iam/src/utils/logging.ts
+++ b/iam/src/utils/logging.ts
@@ -1,0 +1,18 @@
+import winston, { createLogger, format, transports } from "winston";
+
+export const LOG_FORMAT = format.printf(({ level, message, label, timestamp }) => {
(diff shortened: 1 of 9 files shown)


shavinac (approved):

iam/src/utils/logging.ts
```diff
@@ -0,0 +1,18 @@
+import winston, { createLogger, format, transports } from "winston";
```
shavinac, line 1: is it possible to also silence all `console.log`s (at least in production)? `brightid_sdk` is logging every axios request they make 😒

farque65:
Should there be logging in the procedure functions? This should include logging for Brightid

dhx12:
Going to close this one for now. We can pull Winston in when we determine what structured logs we actually want in Datadog.

PR #117, opened 2022-05-25
fix(iam): enable facebook verification with non-owner/developer accounts

### LINKED ISSUE #116: [BUG] Facebook Stamp is unable to verify with a non-developer Facebook account
Now that our Facebook app is switched to "Live" mode, regular Facebook users can interact with the app without needing to be an owner/developer or explicitly added to "App Testers".

However, the way the IAM server is "verifying" the Facebook account is by inspecting the access token value that is returned after the Facebook log in dialog. In order to inspect the user's token, we will need to generate an _app_ token that has permissions to use the inspection endpoint. Currently the IAM server is using the _user token_ to use the inspection endpoint, which is failing because regular users don't have permission for that.

Fix: generate an app token for the `access_token` parameter (the token with permission to use the inspection endpoint), and only use the user token as the `input_token` parameter (the token to be inspected)

shavinac:
This fixes a bug where Facebook stamp verification fails when used with a Facebook account that is
not associated with the Gitcoin Passport Facebook app development team. 

The root cause is that an app token (or similarly privileged user token from a dev team user) is necessary to access the
Facebook Graph API we are using to inspect details about the /user/ access token provided by the
Facebook login on the front end. Therefore, under normal usage, the app access token and the user
access token are two different token values; the bug is that we were attempting to use the user
access token for both purposes. 

This was not caught during development/testing/initial acceptance
because the developer Facebook account was used, not a 'normal' Facebook account.

[fixes #116]


dhx12 (approved):
LGTM

Issue #108, opened 2022-05-23
[CHORE]  Invalidate / Remove old test passports


yonocan:
**Notes**
Might get resolved with issue #101

Issue #101, opened 2022-05-23
🚗 [GRANTS] Hashing on merkel root property


yonocan:
**Notes**
Hashing on merkel root property: change property name on the vc, redeploy to ceramic, want to switch from Merkel root, schema changes but no functional changes in the backend

**Dev notes**
- Schema changes: `root` changes to `hash` in VerifiableCredential
- Root value changes to - `hash: SHA256(IAM_PRIVATE_KEY + VERSION + TYPE + PII_FIELD(s))`
- Additional schema changes from Ceramic team:
  - Use `did:pkh:eip155:1:<eth-address>` instead of `did:ethr:<eth-address>`
  - Separate `#${record.type}` from `credentialSubject.id` and move to `credentialSubject.provider`

shavinac:
the `credentialSubject.hash` field in a stamp VC (see example below) should be populated. 

previously there was a `credentialSubject.root` that held a merkle tree root holding the account information/PII of the stamp verification. we decided to move away from the merkle tree because it would be possible to associate account info (e.g. email, twitter handle, etc) <-> stamp/passport holder by attempting to reconstruct the merkle tree and checking the computed root value. for example, someone could brute-force the email address value from the VC.
see [sybil stories notion](https://gitcoin.notion.site/gitcoin/Sybil-Stories-a2ddda2bf11c42b3b9659d294e9b1760) for more context

```
{
  "provider": "Facebook",
  "credential": {
    "type": [
      "VerifiableCredential"
    ],
    "proof": {
      "jws": "eyJhbGciOiJFZERTQSIsImNyaXQiOlsiYjY0Il0sImI2NCI6ZmFsc2V9..WWC40qMUkpifwT_IDGuwHKuxEHrJ9euyMbqsqTZWBC-4yCVeG8Yhm6_ck5gCaLAH-KduOruSKDrvx3wJiIasBQ",
      "type": "Ed25519Signature2018",
      "created": "2022-06-10T21:35:34.085Z",
      "proofPurpose": "assertionMethod",
      "verificationMethod": "did:key:z6MkghvGHLobLEdj1bgRLhS4LPGJAvbMA1tn2zcRyqmYU5LC#z6MkghvGHLobLEdj1bgRLhS4LPGJAvbMA1tn2zcRyqmYU5LC"
    },
    "issuer": "did:key:z6MkghvGHLobLEdj1bgRLhS4LPGJAvbMA1tn2zcRyqmYU5LC",
    "@context": [
[...]

PR #121, opened 2022-05-27
Main hash

nutrina:
This implements #101


farque65:
The change from did:ether to did:pkh invalidates old passports and closes #108

PR #148, opened 2022-06-03
test(database-client): run integration tests against community testnet node

shavinac:
- add `database-client/integration-tests/integration-test-model-aliases.json` so integration tests can run against testnet definitions
- change `database-client/src/ceramicClient.ts` constructor to use community node as default `ceramicHost`
- add optional `aliases` parameter in `database-client/src/ceramicClient.ts` constructor
- add integration tests to pre-push hook and github ci workflow

While there are updates to the docker-compose file to run ceramic integration tests, currently this is unnecessary since the tests are now running against an external community node. But in the future it would be nice to be able to run tests against a Dockerized local ceramic node that does not persist data


database-client/integration-tests/run-ceramic-tests.sh
```diff
@@ -3,22 +3,20 @@
...
-yarn workspace @dpopp/schemas create-model
-yarn workspace @dpopp/schemas publish-model
+# === fetch passport and VC definitions ===
+curl $CERAMIC_CLIENT_URL/api/v0/streams/kjzl6cwe1jw145znqlxwwar1crvgsm3wf56vcnxo6bu87fqsi6519eypjnzs7mu
+curl $CERAMIC_CLIENT_URL/api/v0/streams/kjzl6cwe1jw149zuvayqa89nhmlvwm0pkdkj0awlxhmtbbay6i972xuwy14jg4f
 
 # === run ceramic integration tests ===
-yarn test:integration
\ No newline at end of file
+yarn test:ceramic-integration
```
shavinac, line 22: Note: this script is only used as the entrypoint in the docker-compose, and you don't actually need to run docker-compose to run the integration tests. Now that we can just reference Ceramic schemas published to our self-hosted testnet node, there is no need to re-create/re-publish schemas or overwrite files.

So now the integration tests can run against the community node (or our self-hosted node if we want to), and should be safe to run outside of a container (doesn't require overwriting any files)

shavinac, line 22: Note: this script is only used as the entrypoint in the docker-compose, and you don't actually need to run docker-compose to run the integration tests. Now that we can just reference Ceramic schemas published to our self-hosted testnet node, there is no need to re-create/re-publish schemas or overwrite files.

So now the integration tests can run against the community node (or our self-hosted node if we want to), and should be safe to run outside of a container (doesn't require overwriting any files)

farque65 (approved):
LGTM

dhx12 (approved):

Issue #139, opened 2022-06-02
[CHORE]: Front-end logging aggregation


dhx12:
Need to clean up and make front-end logging consistent. We should also look into if we can aggregate the logs somewhere for debugging, e.g. Datadog, etc

#### Datadog
- Need Client Token and Application ID (generated in Datadog)
- RUM Browser Monitoring to track user journeys in the app and through the backend
- Browser Log collection can be attached to user sessions and tied to events on the IAM server

brentmartin:
Things to log:
- successful stamps of each
- failures of each stamp

PR #162, opened 2022-06-07
feat(infra): add Datadog container to IAM ECS deployment

dhx12:
- Add the datadog agent container to the Review environment IAM deployment
- add permissions to IAM role to allow listing/describing containers/clusters

[#139]


dhx12:
Seems like we have sufficient backend logging not in order to triage issues, so I am closing this unless we identify a need to add this.

PR #186, opened 2022-06-09
build: rename package scopes @dpopp -> @gitcoin

shavinac:
addresses a security vulnerability due to the @dpopp scope, which is not owned
by gitcoin. also helps remove dpopp mentions in favor of gitcoin passport

also rename `identity/readme.md` -> `identity/README.md`


shavinac:
> When attempting to run tests or start this locally I see the following issue "Cannot find module '@gitcoin/passport-identity/dist/commonjs' or its corresponding type declarations"

yeah, it's been causing the CI tests to fail. for some reason it doesn't happen locally for me 🤔 but i'm looking into it

iam/src/index.ts
```diff
@@ -20,11 +20,15 @@ import {
...
+} from "@gitcoin/passport-types";
 
 // ---- Generate & Verify methods
 import * as DIDKit from "@spruceid/didkit-wasm-node";
-import { issueChallengeCredential, issueHashedCredential, verifyCredential } from "@dpopp/identity/dist/commonjs";
+import {
+  issueChallengeCredential,
+  issueHashedCredential,
+  verifyCredential,
+} from "@gitcoin/passport-identity/dist/commonjs";
```
farque65, line 31:
```suggestion
} from "@gitcoin/passport-identity/dist/commonjs/src/credentials";
```
Should this import also be updated?

shavinac: updating this import to `commonjs/src/credentials` here and in the test file worked on a fresh git clone 👍 , commit amended

farque65 (approved):

PR #192, opened 2022-06-10
GitHub stamp

nutrina:
This implements the github stamp.

In order to get the Github client id and access key, see: https://docs.github.com/en/developers/apps/building-oauth-apps/creating-an-oauth-app


iam/src/procedures/githubOauth.ts
```diff
@@ -0,0 +1,54 @@
+import crypto from "crypto";
+import axios from "axios";
+
+export const getSessionKey = (): string => {
```
shavinac, line 4: is this session key needed? i don't see a need to maintain state over a verification "session", unless you add in a check to ensure the "session key" generated here, matches the `state` passed back from the github redirect.

otherwise it seems like everything here could be moved into the `providers/github.ts` file, since it's only used during the `verifyGithub` process.

iam/src/procedures/index.ts
```diff
@@ -10,19 +11,40 @@ export type GenerateTwitterAuthUrlRequestBody = {
...
+router.post("/github/generateAuthUrl", (req: Request, res: Response): void => {
+  const { callback } = req.body as GenerateGithubAuthUrlRequestBody;
+  if (callback) {
+    const state = githubOAuth.getSessionKey();
+    const clientId = process.env.GITHUB_CLIENT_ID;
 
     const data = {
       state,
-      authUrl: generateAuthURL(client, state),
+      authUrl: `https://github.com/login/oauth/authorize?client_id=${clientId}&redirect_uri=${callback}&state=${state}`,
```
shavinac, line 47: this URL could be generated entirely on the front-end; the query parameters are all public info since the `client_secret` isn't required here. 
maybe even the `state` could be generated & checked on the front-end since the oauth redirect opens a new window, and the current page should(?) still have the initial `state` from the auth url

iam/src/providers/github.ts
```diff
@@ -0,0 +1,46 @@
...
+    return {
+      valid: valid,
+      record: {
+        id: verifiedPayload.id,
+      },
+    };
+  }
+}
+
+// Perform verification on twitter access token
```
shavinac, line 43: np: `twitter access` -> `GitHub access`

app/components/CardList.tsx
```diff
@@ -15,6 +16,7 @@ export const CardList = (): JSX.Element => {
         <PoapCard />
         <EnsCard />
         <PohCard />
+        <GithubCard />
```
farque65, line 19: Minor change: This should be moved before BrightidCard so the web2 identity providers are grouped together.

iam/src/providers/github.ts
nutrina: Fixed

app/components/CardList.tsx
nutrina: Done

iam/src/procedures/githubOauth.ts
nutrina: The session key was not needed (that was used as the state parameter). The creation & verification of state parameter has been move to front-end.
I have removed the filed. The functions that are still required were moved to  `providers/github.ts`.

app/pages/index.tsx
```diff
@@ -59,6 +59,20 @@ const App: NextPage = () => {
 
     return <div></div>;
   }
+  // if Twitter oauth then submit message to other windows and close self
```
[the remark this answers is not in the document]
nutrina: Done

iam/src/procedures/index.ts
```diff
@@ -10,19 +10,23 @@ export type GenerateTwitterAuthUrlRequestBody = {
   callback: string;
 };
 
+export type GenerateGithubAuthUrlRequestBody = {
```
shavinac, line 13: nitpick: unused type

shavinac (approved):

iam/.env-example.env
```diff
@@ -6,6 +6,8 @@ TWITTER_CLIENT_SECRET=MY_APP_SECRET
 FACEBOOK_APP_ID=MY_APP_ID
 FACEBOOK_APP_SECRET=MY_APP_SECRET
 BRIGHTID_PRIVATE_KEY=BRIGHTID_PRIVATE_KEY
+GITHUB_CLIENT_ID=MY_GITHUB_CLIENT_ID
```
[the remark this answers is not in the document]
nutrina: Done

iam/src/procedures/index.ts
nutrina: I have removed that

iam/__tests__/github.test.ts
```diff
@@ -0,0 +1,133 @@
...
+      },
+    } as unknown as RequestPayload);
+
+    expect(githubPayload).toMatchObject({ valid: false });
+  });
+
+  it("should return invalid payload when a bad status code is returned by github user api", async () => {
+    mockedAxios.get.mockImplementation(async (url, config) => {
+      return {
+        status: 500, // This will cause the method
```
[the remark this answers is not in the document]
nutrina: I have removed the comment.

shavinac (approved):

Issue #195, opened 2022-06-13
⭐️ [FEATURE] allow users to download their passport JSON from the passport app


kevinrolsen:
Bypass ceramic and allow users to download the JSON. oftheir passport

As a user of passport
When I have a passport
I want to download my passport data
So that I can be the trusted store of my data

GIVEN that I am a passport holder
WHEN I am viewing my passport
THEN I see the option to download my passport
AND WHEN I click that button
THEN my passport is downloaded as JSON

Etovardesign:
@leoneric Figma: https://www.figma.com/file/RIsPkjpIaeEc73SjLiHK2G/Passport?node-id=2565%3A54352

nutrina:
https://github.com/gitcoinco/dPopp/pull/211

nutrina:
This is done, waiting on PR review before merge.

PR #211, opened 2022-06-14
feat(passport): enable the download of passport as JSON file

nutrina:
This PR adds a download button on the modal that shows the passport in JSON format.

[#195](https://github.com/gitcoinco/dPopp/issues/195)

### DIFF
--- a/app/__tests__/components/JsonOutputModal.test.tsx
+++ b/app/__tests__/components/JsonOutputModal.test.tsx
@@ -27,7 +27,7 @@ afterEach(() => {
 });
 
 describe("displays the modal", () => {
-  it("should not display your modals content", () => {
+  it("should display your modals content", () => {
     render(<JsonOutputModal {...props} />);
 
     const subheading = screen.queryByText("Custom subheading");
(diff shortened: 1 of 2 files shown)


app/__tests__/components/JsonOutputModal.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+    closeButtonText: "Custom Close",
+  };
+});
+
+afterEach(() => {
+  jest.clearAllMocks();
+});
+
+describe("displays the modal", () => {
+  it("should not display your modals content", () => {
```
farque65, line 30:
```suggestion
  it("should display your modals content", () => {
```
This test is checking for the content to be displayed. The title should indicate that.

shavinac (approved):
this still requires ceramic to be up and running in order to fetch the passport in the first place, correct? 
#195 mentions 
> Bypass ceramic and allow users to download the JSON [of their] passport


but this is not really outlined in the ACs 🤔

nutrina:
> 

Yes, that is right.

Issue #241, opened 2022-06-16
[CHORE] Explicitly pin all passport streamids


shavinac:
addressing user data issue where existing passports are not being pinned. any unpinned streams will disappear when the node is restarted (if not sooner when they age out of the cache), which could lead to anchor failures and/or data loss when passport is updated

we could add code to explicitly pin passport streamid after fetching/reading, such as using `ceramicClient.pin.add(passport StreamID)`

we could also use an indexer like https://cerscan.com/mainnet/family/kjzl6cwe1jw148h1e14jb5fkf55xmqhmyorp29r9cq356c7ou74ulowf8czjlzs to view passport activity, and pin as many as we can using `POST /api/v0/pins/:streamid` to our ceramic node

shavinac:
Shavina: currently running a script to pin the first 1000 passports found under cerscan

[streams-to-pin.txt](https://github.com/gitcoinco/dPopp/files/8923050/streams-to-pin.txt)

shavinac:
Fixed by https://github.com/gitcoinco/dPopp/pull/242

PR #242, opened 2022-06-17
fix(database-client): pin passport streams when fetched and when adding a stamp

shavinac:
fix(database-client): pin passport streams when fetched and when adding a stamp

we discovered an issue where passport streams were not being pinned when they were created. ceramic
team is already working on a fix for that issue, that will address newly-created passports going
forward. this fix will address existing passports that were already created without being pinned
previously

[#241]


dhx12 (approved):
LGTM ✅

Issue #252, opened 2022-06-20
[BUG] Inconsistent DID creation (some addresses are checksummed and some are not)


gdixon:
The reading side of Passport will always use a lowercase version of the users address to read from, we should make sure we aways writing to a lowercase version of the DID too.

PR #253, opened 2022-06-20
fix: ensures addresses fed to Ceramic are lowercase

gdixon:
Ensures that we feed a lowercase Address in to Ceramic so that the resultant DID is lowercase - this will ensure the DID we read from is always the same as the DID we write to.

Fixes: #252


kevinrolsen (approved):
lgtm

nutrina:
LGTM

PR #277, opened 2022-06-28
Feat (Hackathon): Add GoodDollar identity 

sirpy:
- integrated Login With GoodDollar API to connect existing verified G$ wallet
- integrated with GoodDollar FaceVerification for adhoc verification

The hashed record value is:
```
{
  address: <passport wallet address>,
  whitelistedAddress: <the G$ wallet whitelisted address>
}
```
Any consumer must always verify that whitelistedAddress is still whitelisted and verify its uniqueness in its own dataset if it wants to prevent duplicate users. (see our login-sdk and sdks how to interact with our contracts. https://github.com/GoodDollar/GoodWeb3-mono)

The whitelistedAddress will either be the G$ wallet address if the passport owner generated it using "Login with GoodDollar" or the same as his passport address if he chose "Get Verified" for adhoc verification.

About GoodDollar Identity
https://medium.com/gooddollar/gooddollar-identity-pillar-balancing-identity-and-privacy-part-i-face-matching-d6864bcebf54

Login with GoodDollar Demo
https://user-images.githubusercontent.com/671095/176608994-a7b7881e-2bb1-49c1-987d-2a93a1cf7f0d.mp4

Get Verified Demo

https://user-images.githubusercontent.com/671095/176694501-89abe394-c046-4099-a397-8d185e5848eb.mp4


sirpy:
@leoneric How can we get this approved for running workflows and get it on the testing environment?

farque65:
@sirpy After resolving the merge conflicts and running the stamp provider locally I have the following feedback:
- The sign up process for Good Dollar should not occur in Passport. It would be better to add a link to send a user to a Good Dollar page to complete the full sign up process.
- The phone number verification seems to have issues sending a verification code. I attempted to verify a number several times with no success.
- The changes for this stamp included modifications to the dashboard view. Those changes are not necessary. To make sure Passport remains modular and scalable views like Dashboard should not have logic specific for any stamp. It is best to isolate stamp logic to the card in the front end and to providers/procedures in the IAM.

sirpy:
@farque65 ok thanks, just to clarify.
There are two flows implemented, so it would help if you explain to which one are you referring
1. login with gooddollar - if you have an existing account,this flow is similar to to when you connect any social account.
if you dont have an account you see a login/signup option. this is the same for google twitter etc...
2. verify identity - this is a "custom" flow. where you can verify your uniqueness without opening a G$ wallet/account, and just using your web3 wallet that you connected with to passport.

A. So when you say signup process are you referring to flow 1 or 2?
B. For Phone verification type any code , the default env vars point to our dev env which do not require real code verification and will also always pass faceverification (ie doesnt check for dups). For production this will need to be point to our production env (.env variables already in place) 
C. Regarding dashboard, some changes are required for completing the "redirect flow" because we do not use a popup window to open our app but rather redirect and then redirect back. This code:
```
// Route user to home when wallet is disconnected
  useEffect(() => {
    //this is a fix for nextjs SSR on gooddollar back redirect not to redirect to home, but stay on dashboard
    if (router?.isReady === false) return;
    const gooddollarLogin = searchParams.get("login") || searchParams.get("verified");
    // Gooddollar verified login is added to the path, so stay here we
    if (gooddollarLogin) return;
    else if (!wallet) {
      navigate("/");
    }
  }, [wallet, router]);
  ```
  The GoodDollarWeb3Provider I can probably move inside the card

sirpy:
@farque65 ping

sirpy:
@farque65 @leoneric i'd like to finish this one so it can be merged, we are waiting for a reply...

PR #279, opened 2022-06-29
Bump release v1.0.0

### LINKED ISSUE #109: [CHORE] Create a prod release of v1.0.0
Reminder to set a version on this - for devs for versioning reasons
new name - `passport-app`?

**Notes from IPM**
- Spin up new repository with clean history
- Pull back current repo to private 
- June 6th or June 7th

shavinac:
affects `app`, `database-client`, `iam`, `identity`, `schemas`, `types`
* bump to version 1.0.0

[resolves #109]

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@gitcoin/passport-app",
   "private": true,
-  "version": "0.0.1",
+  "version": "1.0.0",
   "scripts": {
     "start": "next dev",
     "export": "next build && next export",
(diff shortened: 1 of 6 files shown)


shavinac:
> > `database-client` ... `types`
> 
> What do we think about updating and moving these over to the sdk packages before we release `v1.0.0`?

If we consider `v1.0.0` to be ~GR14 version of `dpopp` repo, then I think we should bump the versions now. We can make PRs against `passport-sdk` repo to reflect the newer changes / bugfixes from `dpopp`, but I don't see this as a blocker to bumping `dpopp` -> v1

Also I don't think we _need_ to publish `dpopp` packages to NPM - the `passport-sdk` packages should definitely be published when ready (after updates/bugfixes), but `dpopp` is not meant to be consumed as separate packages. `v1.0.0` is more of a repo tag IMO. 

What are your thoughts?

PR #281, opened 2022-06-30
test(database-client): generate a new random DID for ceramic integration tests

shavinac:
test(database-client): generate a new random DID for ceramic integration tests

Ceramic integration tests are now consistently timing out (and failing). It seems to be due to the
enormous number of commits logged when using the same DID for all test runs that causes tests to
time out. Replacing this with a newly generated DID each time results in fast-enough integration
tests that pass more consistently.

### DIFF
--- a/database-client/integration-tests/ceramicDatabaseTest.ts
+++ b/database-client/integration-tests/ceramicDatabaseTest.ts
@@ -11,10 +11,7 @@ let testDID: DID;
 let ceramicDatabase: CeramicDatabase;
 
 beforeAll(async () => {
-  const TEST_SEED = new Uint8Array([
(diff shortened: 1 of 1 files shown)


gdixon (approved):
LGTM! Nice catch 🚀

Side note: does this mean there is an upper limit to how many writes we can safely make to the same ceramic stream?

shavinac:
> LGTM! Nice catch 🚀
> 
> Side note: does this mean there is an upper limit to how many writes we can safely make to the same ceramic stream?

The DID's passport we were using for our integration tests has ~1800 commits in its commit log, and now consistently times out when attempting to load the stream from a node that doesn't have it ready.

To be fair, the [ceramic docs](https://developers.ceramic.network/docs/advanced/standards/stream-programs/tile-document/) state that TileDocument limitations include that it is a "growing log, not adapted to frequent updates", and 
> Performance – As more updates are made to a single tile, the underlying DAG grows linearly, and so does sync times when fetching the stream over the network. However, when loading a tile from a node that already has it present, responses are very quick.

PR #309, opened 2022-07-12
chore(app): use enum for isLoadingPassport type instead of "boolean | undefined"

### LINKED ISSUE #275: [TECH DEBT] use more expressive loading states
in our app, some state variables (e.g. `isLoadingPassport`) are using `boolean | undefined` where `undefined` is the "third option" meaning error state.

another example is `passport: Passport | undefined | false` where `passport: false` means "no passport" while `passport: undefined` means "don't know if there is a passport or not"

we should use better/clearer types/enums for these instances

shavinac:
chore(app): use enum for isLoadingPassport type instead of "boolean | undefined"

this should improve readability and prevent possible future bugs from boolean logic and the
falsiness of 'undefined'

[resolves #275]


gdixon (approved):
LGTM! Nicely done @shavinac!

Issue #314, opened 2022-07-15
[TECH DEBT] move dpopp workspaces into a `packages/*` directory


erichfi:
* mostly an organizational improvement
* move `app`, `database-client`, `iam`, `identity`, `schemas`, `types`
* leave `infra` folder in root
* with a `packages/*` directory, then yarn workspaces can be shortened to `"workspaces": ["packages/*"]` and lerna packages can be shortened to `"packages": ["packages/*"]`

PR #347, opened 2022-07-25
refactor: moves packages into a dedicated packages dir

gdixon:
This PR refactors the project to move packages a layer deeper (inside the `packages` dir).

--

Closes: [#314](https://github.com/gitcoinco/passport/issues/314)


farque65:
not a blocker: it would be great to add test:app and lint:app commands to the root package.json file

PR #362, opened 2022-07-30
Feat: GitHub org provider  - For Grant Hub and Round Manager

schultztimothy:
Context from open discussion
```
We could add the issuance of a new class of Twitter stamp to the Passport IAM server that we don't highlight in the passport ui. In fact, this wouldn't be a stamp at all, it would just be the process for issuing a new VC. 

All the IAM server is concerned with is validating that the provided oauth association is valid and it will construct and sign a VC with those details in response. Its already abstracted away from the frontend service and doesn't communicate at all with anyone other than the user and the providers/chain.

So, if we are going to go down the road of creating a service to handle this, we already have one ready in Passport - we just need to add the issuance of a plaintext Twitter stamp.

I would also suggest storing the VC on IPFS and storing the pointer onchain, that should cover everything and allow you to verify everything in-house (locally - once you've pulled in the data).
```
These changes would add a new provider that issues a credential asserting that a github account is a member of the passed github organization.

This credential would be stored on ipfs with an on chain pointer. The assumption is that it could then be verified from within the round manager using the passport-sdk

### DIFF
--- a/iam/src/providers/githubOrg.ts
+++ b/iam/src/providers/githubOrg.ts
@@ -1,0 +1,114 @@
+// ----- Types
+import type { RequestPayload, VerifiedPayload } from "@gitcoin/passport-types";
+import type { Provider, ProviderOptions } from "../types";
+import { requestAccessToken } from "./github";
+import axios from "axios";
+
+export type GithubTokenResponse = {
+  access_token: string;
+};
+
+export type GithubFindMyUserResponse = {
+  id?: string;
+  login?: string;
+  type?: string;
+};
+
+type GHUserRequestPayload = RequestPayload & {
+  org?: string;
+};
+
+type GithubMyOrg = {
+  providedOrg: string;
+  matchingOrg: string;
+};
+
+// Export a Github Provider to carry out OAuth and return a record object
+export class GithubOrgProvider implements Provider {
+  // Give the provider a type so that we can select it with a payload
+  type = "GithubOrg";
+
+  // Options can be set here and/or via the constructor
+  _options = {};
+
+  // construct the provider instance with supplied options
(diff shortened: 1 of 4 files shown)


iam/src/providers/githubOrg.ts
```diff
@@ -0,0 +1,114 @@
...
+    } catch (e) {
+      return { valid: false };
+    } finally {
+      valid = validPayload && validPayload.matchingOrg === validPayload.providedOrg;
+    }
+
+    return {
+      valid: valid,
+      record: {
+        org: validPayload.matchingOrg,
```
gdixon, line 55: To make the PII hash more useful it might be an idea to drop the users handle here too. With this as is, if multiple users are attesting to the same org they will have the same hash so we won't be able to spot duplicates

gdixon (changes requested):

iam/src/providers/githubOrg.ts
schultztimothy: Great callout! I just included that

iam/src/providers/githubOrg.ts
```diff
@@ -0,0 +1,124 @@
...
+    } finally {
+      const validOrg = ghVerification?.validOrg;
+      valid = validOrg && validOrg.matchingOrg === validOrg.providedOrg;
+    }
+
+    return {
+      valid: valid,
+      record: {
+        org: ghVerification.validOrg.matchingOrg,
+        handle: ghVerification.handle,
```
gdixon, line 62: Sorry to do this to you @schultztimothy but this should really be `ghVerification.id`, the user can change their handle but not their ID (I'd forgotten about this, @aminah-io reminded me yesterday).

```suggestion
        id: ghVerification.id,
```

Issue #69, opened 2022-05-09
[TECH DEBT] Ceramic CLI dependency causing issues with NextJS export process


shavinac:
Currently if running `next export` to build the web app code, it errors with 
```
info  - Loaded env from /Users/shavinachau/workspace/gitcoinco/dPopp/app/.env
Failed to compile.

../node_modules/@tendermint/belt/src/base64.ts:48:43
Type error: Type 'Uint8Array' is not an array type or a string type. Use compiler option '--downlevelIteration' to allow iterating of iterators.

  46 |  */
  47 | export function bytesToBase64 (bytes: Bytes): Base64String {
> 48 |     const binary = String.fromCharCode(...bytes);
     |                                           ^
  49 |     return btoa(binary);
  50 | }
  51 |
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

This issue is caused by the `@tendermint` family of libraries, which is a dependency of `@ceramicnetwork/cli`, which is brought into the NextJS export process because NextJS is pulling in the **root-level node modules** from lerna bootstrap, which includes dependencies from multiple packages including those that require `@ceramicnetwork/cli`

for now, we hacked the Fleek build command to remove the `@tendermint` package `
npm install --g lerna && lerna bootstrap && rm -rf ../node_modules/@tendermint && npm run export` -- this is super hacky and we should figure out a better way to deal with this

aminah-io:
Perhaps in `lerna.json` you can try adding `nohoist` to `command.bootstrap` after the `packages` array:
```
{
  "packages": [...],
  "command": {
    "bootstrap": {
      "nohoist": [
        "@tendermint",
      ],
    ...
}
```

This _should_ prevent the package from being installed at the repo root. You can also try it in the build command as well, `lerna bootstrap --nohoist=@tendermint`

[Source](https://github.com/lerna/lerna/tree/main/commands/bootstrap)

aminah-io:
#375

PR #375, opened 2022-08-02
fix: added fix for #69 to prevent installation of @tendermint dep during next…

aminah-io:
…js export

#69


gdixon (approved):
Awesome! Thanks @aminah-io 🚀

PR #378, opened 2022-08-03
feat(iam): adds provider to check if user has 5 or more GitHub repos

aminah-io:


iam/src/providers/fiveOrMoreGithubRepos.ts
```diff
@@ -0,0 +1,92 @@
...
+      return { valid: false };
+    } finally {
+      valid = verifiedPayload && verifiedPayload.public_repos >= 5 && verifiedPayload.id ? true : false;
+    }
+
+    return {
+      valid: valid,
+      record: {
+        id: verifiedPayload.id,
+        repoCount: `${verifiedPayload.public_repos}`,
```
gdixon, line 50: It might be better to have this as a statement that matches the range rather than the number of repos that the user has - if the user creates a new repo, they would get a new hash, and would be able to use that new stamp against a different Passport.

gdixon (changes requested):
Just the one tiny comment - the rest looks awesome! Nice work @aminah-io 🚀

gdixon (approved):
LGTM!! Nicely done 🚀

farque65 (approved):
LGTM once the merge conflicts are resolved

PR #379, opened 2022-08-03
feat(iam): adds twitter follower stamp provider

farque65:
Following Stamps are included:
>100, >500, >=1000, >5000

Closes #363
Closes #365
Closes #366
Closes #370

PR #380, opened 2022-08-04
feat(iam): adds twitter greater than 10 tweets/posts stamp provider

### LINKED ISSUE #371: [IAM] As an IAM server, I want to issue stamps for >10 Tweets/Posts on Twitter
GIVEN I'm the Gitcoin IAM server,
AND Passport holder wants to verify their number of Twitter posts
WHEN I run them through the Twitter oAuth process
THEN I call the Twitter API for their number of posts/tweets
AND WHEN their account has >=10 Twitter posts
THEN I issue them a VC that verifies that they have >=10 Twitter posts

**Notes:**
- https://developer.twitter.com/en/docs/api-reference-index

farque65:
Closes #371

### DIFF
--- a/iam/src/providers/twitterTweets.ts
+++ b/iam/src/providers/twitterTweets.ts
@@ -2,15 +2,15 @@
 import type { RequestPayload, VerifiedPayload } from "@gitcoin/passport-types";
 
 // ----- Twitters OAuth2 library
-import { deleteClient, getClient, getTweetCount } from "../procedures/twitterOauth";
+import { deleteClient, getClient, getTweetCount, TwitterTweetResponse } from "../procedures/twitterOauth";
 import type { Provider, ProviderOptions } from "../types";
 
 // Perform verification on twitter access token and retrieve follower count
-async function verifyTwitterTweets(sessionKey: string, code: string): Promise<number> {
+async function verifyTwitterTweets(sessionKey: string, code: string): Promise<TwitterTweetResponse> {
   const client = getClient(sessionKey);
   const data = await getTweetCount(client, code);
   deleteClient(sessionKey);
-  return data.tweetCount;
+  return data;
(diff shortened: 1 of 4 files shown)


iam/src/providers/twitterTweets.ts
```diff
@@ -0,0 +1,48 @@
...
+    } catch (e) {
+      return { valid: false };
+    } finally {
+      valid = tweetCount > 10;
+    }
+
+    return {
+      valid: valid,
+      record: {
+        tweetCount: tweetCount?.toString(),
```
gdixon, line 44: We should avoid putting counters as attestation and should favour statements instead, if a user gains a new follower, they could produce this same stamp with a different hash.

Instead we should consider doing something like: followerCount: valid ? "gt10" : false

gdixon (changes requested):
Same comment on this PR too (as on #379)

iam/src/providers/twitterTweets.ts
gdixon: We should avoid putting counters as attestation and should favour statements instead, if a user gains a new follower, they could produce this same stamp with a different hash.

Instead we should consider doing something like: `tweetCount: valid ? "gt10" : false`

iam/src/providers/twitterTweets.ts
```diff
@@ -0,0 +1,48 @@
...
+    } catch (e) {
+      return { valid: false };
+    } finally {
+      valid = tweetCount > 10;
+    }
+
+    return {
+      valid: valid,
+      record: {
+        tweetCount: valid ? "gt10" : "",
```
gdixon, line 44: We should also be returning the users twitter handle here

gdixon (approved):
Nicely done!

Issue #368, opened 2022-08-02
[IAM] As an IAM server, I want to issue stamps for whether somebody starred one of their Github repos


leoneric:
GIVEN I'm the Gitcoin IAM server,
AND Passport holder wants to verify that they own a Github repo that somebody starred
WHEN I run them through the Github oAuth process
THEN I call the Github API to see whether they have repo that was starred by somebody
AND WHEN they own a repo that somebody starred
THEN I issue them a VC that verifies that they have a repo that somebody starred

Notes:
- https://docs.github.com/en/rest

Issue #367, opened 2022-08-02
[IAM] As an IAM server, I want to issue stamps for whether somebody forked one of their Github repos


leoneric:
GIVEN I'm the Gitcoin IAM server,
AND Passport holder wants to verify that they own a Github repo that somebody forked
WHEN I run them through the Github oAuth process
THEN I call the Github API to see whether they have repo that was forked by somebody
AND WHEN they own a repo that somebody forked
THEN I issue them a VC that verifies that they have a repo that somebody forked

**Notes:**
- https://docs.github.com/en/rest

PR #392, opened 2022-08-05
feat(iam): #367 stamps for stars & #368 stamps for forks

aminah-io:


aminah-io:
Just had an epiphany--need to add checks for if the user has no repos and also add some to the tests as well.

iam/src/providers/githubForkedRepoProvider.ts
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  // verify that the proof object contains valid === "true"
+  async verify(payload: RequestPayload): Promise<VerifiedPayload> {
+    let valid = false,
+      verifiedUserPayload: GithubFindMyUserResponse = {},
+      verifiedUserRepoPayload: GithubUserRepoResponseData = {};
+
+    try {
+      verifiedUserPayload = await verifyGithub(payload.proofs.code);
```
nutrina, line 37: I would recommend requesting the access token at the beginning of this block, and passing it as a parameter to the `verify*` function.
Then these functions do not need to request a token, and we save 1 request to github.

iam/src/providers/githubStarredRepoProvider.ts
```diff
@@ -0,0 +1,151 @@
...
+
+  // verify that the proof object contains valid === "true"
+  async verify(payload: RequestPayload): Promise<VerifiedPayload> {
+    let valid = false,
+      verifiedUserPayload: GithubFindMyUserResponse = {},
+      verifiedUserRepoPayload: GithubUserRepoResponseData = {};
+
+    try {
+      verifiedUserPayload = await verifyGithub(payload.proofs.code);
+      verifiedUserRepoPayload = await verifyUserGithubRepo(verifiedUserPayload, payload.proofs.code);
```
nutrina, line 38: Same observation as above: get the access token once, and pass it in as argument to both `verify*` functions

nutrina (changes requested):
I would try to reduce the number of `requestAccessToken` requests

iam/src/providers/githubStarredRepoProvider.ts
```diff
@@ -0,0 +1,151 @@
...
+          // Return the first instance where the stargazer type is a "User"
+          const stargazersItem: Record<string, unknown> = stargazerData.find(
+            (stargazerObject: GithubUserRepoResponseData["owner"]) => {
+              stargazerObject.type === userData.type;
+            }
+          );
+          if (stargazersItem.id !== userData.id) {
+            return repo;
+          }
+        };
```
nutrina, line 142: This is an anonymous function that never gets called, hence the code in it never gets called.

iam/__tests__/githubStarredRepoProvider.test.ts
```diff
@@ -0,0 +1,236 @@
...
+    });
+  });
+
+  it("should return invalid payload when unable to retrieve auth token", async () => {
+    mockedAxios.post.mockImplementation(async (url, data, config) => {
+      return {
+        status: 500,
+      };
+    });
+
```
nutrina, line 105: In all cases where we expect a negative verify, we should check that the axios mock was called as expected. Because otherwise we have no certainty if the expected checks have been performed before retuning `valid: false`

iam/src/providers/githubStarredRepoProvider.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    try {
+      accessToken = await requestAccessToken(payload.proofs.code);
+      verifiedUserPayload = await verifyGithub(accessToken);
+      verifiedUserRepoPayload = await verifyUserGithubRepo(verifiedUserPayload, accessToken);
+    } catch (e) {
+      return { valid: false };
+    } finally {
+      valid = verifiedUserPayload && verifiedUserPayload.id && verifiedUserRepoPayload ? true : false;
+    }
```
nutrina, line 49: I would change this:

```suggestion
    try {
      accessToken = await requestAccessToken(payload.proofs.code);
      verifiedUserPayload = await verifyGithub(accessToken);
      verifiedUserRepoPayload = await verifyUserGithubRepo(verifiedUserPayload, accessToken);

      valid = verifiedUserPayload && verifiedUserPayload.id && verifiedUserRepoPayload ? true : false;
    } catch (e) {
      return { valid: false };
    } 
```

iam/src/providers/githubStarredRepoProvider.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    return {
+      valid: valid,
+      record: {
+        id: `${verifiedUserPayload.id}gte1Star`,
+      },
+    };
```
nutrina, line 56: And also this:
```suggestion
    return {
      valid: valid,
      record: valid ? {
        id: `${verifiedUserPayload.id}gte1Star`,
      } : undefined,
    };
```

iam/src/providers/githubStarredRepoProvider.ts
```diff
@@ -0,0 +1,143 @@
...
+          const stargazerResponseData: [] = stargazerResponse.data;
+          return stargazerResponseData;
+        };
+
+        const stargazer = await stargazerCheck();
+        for (let i = 0; i < stargazer.length; i++) {
+          stargazerData = stargazer[i];
+        }
+
+        return stargazerData.id === repo.owner.id ? false : repo;
```
nutrina, line 135: Also I would simplify this:

```suggestion
        // GET the solo stargazer's data
        const stargazer = (await axios.get(repo.stargazers_url)).data;

        // Length should actually be 1 ... but we iterate just in case another star was added in the mean time
        for (let i = 0; i < stargazer.length; i++) {
          if (stargazer[i].id != repo.owner.id) {
            return true;
          }
        }
        return false;        
```

iam/src/providers/githubForkedRepoProvider.ts
```diff
@@ -0,0 +1,126 @@
...
+  // Returns an object containing first instance of a user's repo if it has been forked,
+  // or the last checked repo with no forks
+  const userRepoForksCheck = repoRequest.data.find((repo: GithubUserRepoResponseData): GithubUserRepoResponseData => {
+    // Check to see if the authenticated GH user is the same as the repo owner,
+    // if the repo is not a fork of another repo, and if the repo fork count is gte 1
+    if (userData.id === repo.owner.id && !repo.fork && repo.forks_count >= 1) {
+      return repo;
+    } else {
+      return repo;
+    }
```
nutrina, line 122: We have the same return on the `if` and `else` branch. Returning `true` in he if branch and `false` by default (or on the else branch) would solve the issue.

iam/__tests__/githubForkedRepoProvider.test.ts
```diff
@@ -0,0 +1,223 @@
...
+    const forkedGithubRepoProvider = new ForkedGithubRepoProvider();
+
+    const forkedGithubRepoProviderPayload = await forkedGithubRepoProvider.verify({
+      proofs: {
+        code,
+      },
+    } as unknown as RequestPayload);
+
+    expect(forkedGithubRepoProviderPayload).toMatchObject({ valid: false });
+  });
```
nutrina, line 182: This test for example (and also the others) are missing the checks if the proper AJAX requests have been made:

```suggestion
  it.only("should return invalid payload when the fork count for all repos is less than 1", async () => {
    mockedAxios.get.mockImplementation(async (url, config) => {
      return {
        data: [
          {
            owner: {
              id: "18723656",
              type: "User",
            },
            fork: false,
            forks_count: 0,
          },
          {
            owner: {
              id: "18723656",
              type: "User",
            },
            fork: true,
            forks_count: 0,
          },
        ],
        status: 200,
      };
    });

    const forkedGithubRepoProvider = new ForkedGithubRepoProvider();

    const forkedGithubRepoProviderPayload = await forkedGithubRepoProvider.verify({
      proofs: {
        code,
      },
    } as unknown as RequestPayload);

    expect(mockedAxios.get).toHaveBeenCalledTimes(2);
    expect(mockedAxios.get).toBeCalledWith(`https://api.github.com/users/${validGithubUserResponse.data.login}/repos`, {
      headers: { Authorization: `token ${validCodeResponse.data.access_token}` },
    });

    expect(forkedGithubRepoProviderPayload).toMatchObject({ valid: false });
  });
```

nutrina (changes requested):
See the comments regarding the topics I have found.

iam/__tests__/githubForkedRepoProvider.test.ts
```diff
@@ -0,0 +1,243 @@
...
+      return validGithubUserResponse;
+    }
+    if (url.endsWith("/repos")) {
+      return validGithubUserRepoResponse;
+    }
+  });
+});
+
+describe("Attempt verification", function () {
+  it("handles valid verification attempt", async () => {
```
nutrina, line 85: Also please create tests for the edge cases:
  - a user has only a repos with fork_count=1 and he is the owner of the fork
  - a user has only a repos with fork_count=1 and he is not owner of the fork

nutrina (changes requested):

iam/__tests__/githubForkedRepoProvider.test.ts
aminah-io: For the forks, a user can't fork their own repo 🙌

nutrina (approved):
LGTM

PR #393, opened 2022-08-05
Twitter clear text

schultztimothy:
This returns the users twitter handle from the new provider and concats it to the provider type so that it can be hashed into the VC

### DIFF
--- a/iam/src/index.ts
+++ b/iam/src/index.ts
@@ -55,6 +55,7 @@ import {
   TwitterFollowerGTE1000Provider,
   TwitterFollowerGT5000Provider,
 } from "./providers/TwitterFollower";
+import { ClearTextGithubOrgProvider } from "./providers/clearTextGithubOrg";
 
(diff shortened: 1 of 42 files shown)


iam/src/index.ts
```diff
@@ -184,7 +188,7 @@ app.post("/api/v0.0.0/verify", (req: Request, res: Response): void => {
                 // construct a set of Proofs to issue a credential against (this record will be used to generate a sha256 hash of any associated PII)
                 const record: ProofRecord = {
                   // type and address will always be known and can be obtained from the resultant credential
-                  type: payload.type,
+                  type: verifiedPayload.record.pii ? `${payload.type}#${verifiedPayload.record.pii}` : payload.type,
```
gdixon, line 191: ❤️‍🔥

gdixon (approved):
This is awesome @schultztimothy - thank you 🙏

PR #394, opened 2022-08-05
GitHub org clear text provider

schultztimothy:
This builds off of https://github.com/gitcoinco/passport/pull/393 to return a cleartext VC validating that a user belongs to given github organzation


gdixon (approved):
LGTM! 🚀

Issue #377, opened 2022-08-03
[IAM] Verify multiple VCs with one signature and oAuth flow


leoneric:

nutrina:
This has been verified and tested.

PR #396, opened 2022-08-08
feat: allows for user to request multiple VCs in single req

gdixon:
This PR refactors to allow for multiple VCs to be requested and issued from the IAM server.

--

refers: #377


nutrina (approved):
LGTM

Issue #369, opened 2022-08-02
[IAM] As an IAM server, I want to issue stamps for whether somebody has a Facebook profile picture


leoneric:
GIVEN I'm the Gitcoin IAM server,
AND Passport holder wants to verify that they have a Facebook profile picture
WHEN I run them through the Facebook oAuth process
THEN I call the Facebook API to see whether they have a Facebook profile picture
AND WHEN they have a profile picture
THEN I issue them a VC that verifies that they have a Facebook profile picture

nutrina:
PR is open #397

PR #397, opened 2022-08-09
Facebook friends

### LINKED ISSUE #353: [IAM] As an IAM server, I want to issue stamps for  >=100 Facebook friends
GIVEN I'm the Gitcoin IAM server,
AND Passport holder wants to verify their number of Facebook friends
WHEN I run them through the Facebook oAuth process
THEN I call the Facebook API for their number of friends
AND WHEN their account has >= 100 Facebook friends
THEN I issue them a VC that verifies that they have >= 100 Facebook friends

**Notes**
- https://developers.facebook.com/docs/graph-api/reference

nutrina:
This closes #353 and #369

### DIFF
--- a/iam/src/providers/facebookProfilePicture.ts
+++ b/iam/src/providers/facebookProfilePicture.ts
@@ -24,7 +24,7 @@ type Response = {
   };
 };
 
-// Query Facebook graph api to verify the access token recieved from the user login is valid
+// Query Facebook graph api to verify the profile picture
 export class FacebookProfilePictureProvider implements Provider {
   // Give the provider a type so that we can select it with a payload
   type = "FacebookProfilePicture";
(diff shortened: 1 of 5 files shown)


iam/src/providers/facebookProfilePicture.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  // construct the provider instance with supplied options
+  constructor(options: ProviderOptions = {}) {
+    this._options = { ...this._options, ...options };
+  }
+
+  // verify that the proof object contains valid === "true"
+  async verify(payload: RequestPayload): Promise<VerifiedPayload> {
+    try {
+      // Get the FB profile
```
nutrina, line 42: @gdixon Do you think we should validate the `payload.proofs.accessToken` similar like we do in `facebookFriends.ts`?
There is no real reason to do this for this stamp, except for additional security, but I am not sure about this.

aminah-io (commented):
Everything looks good to me!

iam/src/providers/facebookProfilePicture.ts
farque65: After checking out the stamp and looking at the additional access token verification I think it is good to keep it. There does not seem to be a performance impact from it

farque65 (approved):

PR #416, opened 2022-08-12
feat(iam): adds  Self/Community identity staking Stamps

farque65:
For more info on the identity staking app visit [the notion](https://www.notion.so/gitcoin/Identity-Staking-Documentation-0adeb13909184a739deb58d1e3b77043)


iam/src/providers/communitystaking.ts
```diff
@@ -0,0 +1,201 @@
...
+              address: stakeData.address,
+              // csgt100 = Community Staking Greater than 100
+              stakeAmount: "csgt100",
+            }
+          : {},
+      };
+    } catch (e) {
+      return {
+        valid: false,
+        error: [JSON.stringify(e)],
```
nutrina, line 157: As a general rule, I would propose not to simply log or return exceptions.
In this particular example it is not critical.

But in case this would be an axios exception, and if the request we made would contain a secret in the request headers, then this might result in being leaked ...

iam/src/providers/communitystaking.ts
```diff
@@ -0,0 +1,201 @@
...
+              address: stakeData.address,
+              // csgt500 = Community Staking Greater than 500
+              stakeAmount: "csgt500",
+            }
+          : {},
+      };
+    } catch (e) {
+      return {
+        valid: false,
+        error: [JSON.stringify(e)],
```
nutrina, line 197: Same here, please do not return exceptions ...

iam/src/providers/communitystaking.ts
```diff
@@ -0,0 +1,201 @@
...
+              address: stakeData.address,
+              // csgt10 = Community Staking Greater than 10
+              stakeAmount: "csgt10",
+            }
+          : {},
+      };
+    } catch (e) {
+      return {
+        valid: false,
+        error: [JSON.stringify(e)],
```
nutrina, line 116: Same here, please do not return exceptions ...

iam/src/providers/selfstaking.ts
```diff
@@ -0,0 +1,202 @@
...
+              address: payload.address,
+              // ssgt1 = Self Staking Greater than 1
+              stakeAmount: "csgt1",
+            }
+          : {},
+      };
+    } catch (e) {
+      return {
+        valid: false,
+        error: [JSON.stringify(e)],
```
nutrina, line 116: Same here, please do not return exceptions ...

iam/src/providers/selfstaking.ts
```diff
@@ -0,0 +1,202 @@
...
+              address: payload.address,
+              // ssgt5 = Self Staking Greater than 5
+              stakeAmount: "csgt5",
+            }
+          : {},
+      };
+    } catch (e) {
+      return {
+        valid: false,
+        error: [JSON.stringify(e)],
```
nutrina, line 157: Same here, please do not return exceptions ...

iam/src/providers/selfstaking.ts
```diff
@@ -0,0 +1,202 @@
...
+              address: payload.address,
+              // ssgt50 = Self Staking Greater than 50
+              stakeAmount: "csgt50",
+            }
+          : {},
+      };
+    } catch (e) {
+      return {
+        valid: false,
+        error: [JSON.stringify(e)],
```
nutrina, line 198: Same here, please do not return exceptions ...

iam/__tests__/communitystaking.test.ts
```diff
@@ -0,0 +1,248 @@
...
+      users: [{}],
+    },
+  },
+};
+
+interface RequestData {
+  query: string;
+}
+
+describe("Attempt verification", function () {
```
nutrina, line 54: I would also add a test in this suite for: `handles invalid verification attempt where an exception is thrown` because axios seems to throw on errors.

nutrina (changes requested):
I have a few comments.
In addition to those: please rename the files to camel case (as this is the convention used everywhere):
- `comunitystaking.ts` => `comunityStaking.ts`
- `selfstaking.ts` => `selfStaking.ts`
- ...

nutrina (approved):
LGTM

PR #425, opened 2022-08-15
chore(iam): allow mulitple github credentials

schultztimothy:
Allow multiple credentials to be used for `clearTextGithubOrg` provider to account for different clients


iam/__tests__/clearTextGithubOrg.test.ts
```diff
@@ -73,15 +75,16 @@ beforeEach(() => {
 describe("Attempt verification", function () {
   const pii = `${org}#${validGithubUserResponse.data.id}`
   it("handles valid verification attempt", async () => {
```
nutrina, line 77: I would add one more test:
```
it("handles valid verification attempt when no requestedClient is provided", ...
```
and ensure that the correct credentials are used in that case  (these credentials `process.env.GITHUB_*` )

nutrina (changes requested):

iam/__tests__/clearTextGithubOrg.test.ts
schultztimothy: Good call :) I added an additional test to cover this

nutrina (approved):
LGTM

PR #431, opened 2022-08-18
feat(iam): adds snapshot stamps for >=1 proposals voted on, >=2 votes…

aminah-io:
… on proposals, and tests

### DIFF
--- a/iam/__tests__/snapshotProposalsProvider.test.ts
+++ b/iam/__tests__/snapshotProposalsProvider.test.ts
@@ -174,4 +174,21 @@ describe("Attempt verification", function () {
     expect(axios.post).toHaveBeenCalledTimes(1);
     expect(verifiedPayload).toMatchObject({ valid: false });
   });
+
+  it("should return invalid payload when an exception is thrown when a request is made", async () => {
+    mockedAxios.post.mockImplementation(async (url, data) => {
+      const query: string = (data as RequestData).query;
+      if (url === snapshotGraphQLDatabase && query.includes(BAD_MOCK_ADDRESS_LOWER)) {
+        throw "an error";
+      }
+    });
+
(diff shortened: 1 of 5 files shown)


iam/src/providers/snapshotVotesProvider.ts
```diff
@@ -0,0 +1,117 @@
...
+
+  // Options can be set here and/or via the constructor
+  _options = {};
+
+  // construct the provider instance with supplied options
+  constructor(options: ProviderOptions = {}) {
+    this._options = { ...this._options, ...options };
+  }
+
+  // Verify that the address that is passed in owns at least one POAP older than 15 days
```
nutrina, line 47: That is not the POAP verification ...

iam/__tests__/snapshotProposalsProvider.test.ts
```diff
@@ -0,0 +1,177 @@
...
+      proposals: null as [],
+    },
+  },
+};
+
+interface RequestData {
+  query: string;
+}
+
+describe("Attempt verification", function () {
```
nutrina, line 67: I would also add a test like:

`it("should return invalid payload an exception is thrown when request is made",`

iam/__tests__/snapshotVotesProvider.test.ts
```diff
@@ -0,0 +1,203 @@
...
+    },
+  },
+};
+
+interface RequestData {
+  query: string;
+}
+
+describe("Attempt verification", function () {
+  beforeEach(() => {
```
nutrina, line 94: Same as above, I would also add a test like:

`it("should return invalid payload an exception is thrown when request is made",`

nutrina (changes requested):
I have only 3 minor topics

Other than the, nicely done 👍

nutrina (approved):
LGTM

PR #476, opened 2022-08-23
feat(infra): gh action for deploying ceramic to production

nutrina:
This action should provision the ceramic node to production.

This closes: #481

### DIFF
--- a/.github/workflows/ceramic-promote-production.yml
+++ b/.github/workflows/ceramic-promote-production.yml
@@ -1,4 +1,4 @@
-name: Deploy Ceramic Node to Production
+name: Deploy Ceramic Mainnet Node
 
 on:
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


.github/workflows/ceramic-promote-production.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          cache: "yarn"
+          cache-dependency-path: infra/ceramic-mainnet/package-lock.json
+      # Update the pulumi stack with new image
+      - run: |
+          npm install
+          pulumi stack select -c gitcoin/ceramic/mainnet-node
+          pulumi config -s gitcoin/ceramic/mainnet-node set aws:region us-east-1 --non-interactive
+        working-directory: infra/ceramic-mainnet
+        env:
```
kammerdiener, line 27: I believe there are some additional values we need to pass in.

I know I include `ROUTE_53_ZONE` and `DOMAIN`

kammerdiener (commented):
Just need to check that we are getting the correct ENVs through.

.github/workflows/ceramic-promote-production.yml
nutrina: Those are passed in in the step below, as env vars

kammerdiener (approved):
I didn't read far enough down. LGTM!

Issue #304, opened 2022-07-06
As a passport holder, I want my DID Session to be persisted when I revisit Passport, so that I don't need to re-sign to authorize myself


shavinac:
GIVEN that I am a Gitcoin passport holder
AND I have already signed a message with my wallet to "log in" to Passport
WHEN I revisit Passport within 24 hours
THEN I should still be "logged in"
AND I should not re-sign a message with my wallet to "log in" again

Dev notes:
* ([ceramic discord message](https://discord.com/channels/682786569857662976/963185659638607982/994044410176217260)) did-session@0.1.1 - can serialize/store a session now, can check if a session isExpired or not before making writes
* some initial docs -https://developers.ceramic.network/reference/accounts/did-session/

leoneric:
Here's an example for using these functions: https://developers.ceramic.network/reference/accounts/did-session/#typical-usage-pattern

Using localStorage is just for illustrative purposes though, we should take care of properly securing the storage of sessions.

zachferland:
In the latest release of `@self.id/framework@0.3.3` you can now access the session and then manage it as described in docs above and pass it when connecting/auth.

For example with self.id framework
```ts
...
const [connection, connect, disconnect] = useViewerConnection()
...
// get session string you serialized and stored before, check if still valid (or how much longer)
const sessionStr = ...
const selfid = await connect(new EthereumAuthProvider(window.ethereum, accounts[0]), sessionStr)
...
// get session to serialize and store 
const session = selfid.client.session //or connection.selfID.session
session.serialize()
```

nutrina:
Waiting for feedback from ceramic on this: https://discord.com/channels/682786569857662976/963185659638607982/1011583783067189268

PR #477, opened 2022-08-23
Feat(app): persisting user sessions between reloads of the app

nutrina:
This solves: https://github.com/gitcoinco/passport/issues/304

### DIFF
--- a/app/context/userContext.tsx
+++ b/app/context/userContext.tsx
@@ -10,6 +10,7 @@ import { JsonRpcSigner, Web3Provider } from "@ethersproject/providers";
 // -- Ceramic and Glazed
 import { EthereumAuthProvider } from "@self.id/web";
 import { useViewerConnection } from "@self.id/framework";
+import { DIDSession } from "@glazed/did-session";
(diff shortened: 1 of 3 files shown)


app/context/userContext.tsx
```diff
@@ -10,6 +10,7 @@ import { JsonRpcSigner, Web3Provider } from "@ethersproject/providers";
 // -- Ceramic and Glazed
 import { EthereumAuthProvider } from "@self.id/web";
 import { useViewerConnection } from "@self.id/framework";
+import { DIDSession } from "@glazed/did-session";
```
aminah-io, line 13: Noticed you're using `@glazed/did-session` here--was there a specific reason you used it and not `did-session`?

nutrina: That is the library as it is registered with npm, and this is what we use: https://www.npmjs.com/package/@glazed/did-session.
Not sure why the non-prefixed one also exists, as it seems to be the same version: https://www.npmjs.com/package/did-session

PR #478, opened 2022-08-24
feat(iam): adds providers for ETH mainnet transactions, env variables, and infra secrets

aminah-io:
Will ask Leon about the edge cases (count only successful transactions, and transactions FROM the user) during stand-up


aminah-io:
I also need to update and expand the comments! 📝

nutrina (approved):
LGTM.

The only thing I am wondering if we could drop this parameters in the API request.
`startblock=0&endblock=9999999`.
But I do not see in the documentation what the default values would be in this case (or if no block range applies, which is what I would expect).

aminah-io:
> LGTM.
> 
> The only thing I am wondering if we could drop this parameters in the API request. `startblock=0&endblock=9999999`. But I do not see in the documentation what the default values would be in this case (or if no block range applies, which is what I would expect).

Yep, it looks like it can be omitted, so I'll go ahead and drop those

Issue #460, opened 2022-08-23
[STAMP] As a Passport holder, I want to open the Gitcoin Grants sidebar


leoneric:
GIVEN that I’m signed into my Passport
WHEN I click on the Gitcoin Grants stamp
THEN the app opens the Gitcoin Grants sidebar highlighting the data points:
- #398
- #385, #386, #387, #388
- #389, #390
- #399
- #401, #402, #403
- #404, #405, #406
- #407, #408, #409
AND show green light on the left hand side of the already verified data points

Notes:
- Reference designs: https://www.figma.com/file/RIsPkjpIaeEc73SjLiHK2G/Passport?node-id=3752%3A68252
- Reference demo: https://www.loom.com/share/ff733d246c9744a59aeff8ff6a7a435a
- "Gitcoin Grants" stamp image/logo:

![Image](https://user-images.githubusercontent.com/59769856/187163833-69966475-7a01-4e3f-b0dd-47937004b3cb.svg)

PR #533, opened 2022-09-05
feat(app): adds sidebar for Gitcoin platform

gdixon:
This PR exposes all Gitcoin Providers via the Gitcoin Platform sidebar.

--

Closes: #460, closes: #461, closes: #462, closes: #463


types/src/index.d.ts
```diff
@@ -175,6 +176,25 @@ export type PROVIDER_ID =
   | "ForkedGithubRepoProvider"
   | "StarredGithubRepoProvider"
   | "FiveOrMoreGithubRepos"
```
aminah-io, line 178:
```suggestion
  | "FiveOrMoreGithubRepos"
  | "Gitcoin"
```

aminah-io (changes requested):
Just one suggestion, otherwise LGTM!

types/src/index.d.ts
gdixon: We're not issuing a Stamp for Gitcoin on its own (it would be the same as the github stamp), so "Gitcoin" shouldn't need to be a `PROVIDER_ID` I don't think?

aminah-io: > We're not issuing a Stamp for Gitcoin on its own (it would be the same as the github stamp), so "Gitcoin" shouldn't need to be a `PROVIDER_ID` I don't think?

Ahhh yes, that makes sense!

aminah-io (approved):

PR #551, opened 2022-09-07
fix(iam): fix the list of providers for gitcoin stamps

nutrina:
This fixes the errors received when validating the gitcoin stamps due to misconfigured providers.


gdixon (approved):
Nice catch!! 🙌

PR #557, opened 2022-09-07
fix(app): only show updated at if they have previously verified

schultztimothy:
https://user-images.githubusercontent.com/6887938/188927345-1c78e96a-448b-4014-b6b4-022f6164d669.mov


gdixon (approved):
Perfect! 🚀

PR #577, opened 2022-09-12
feat(app): removes brightid verification modal and shows connection modal

### LINKED ISSUE #570: [Bug] Bright ID modal showing verification null
**Scenario**

- After user verifies their Bright ID stamp they see null in the modal

**Expected Result**

- If the verification is unsuccessful they should see the connection modal
- If the verification is successful they should see the done toast and the switch changes to success

**Actual Result**

- After successfully verifying their Bright ID stamp they see the connection modal with a verification modal that shows null

Screen Shot 2022-09-09 at 2 38 06 PM

farque65:
Closes #570


gdixon (approved):
LGTM! 🚀

Issue #567, opened 2022-09-09
[BUG] Passport holders should be able to open a stamp JSON modal on the Passport page showing the stamp‘ VC(s) JSON


gdixon:
**Scenario**
- Logged in Passport holders click on the JSON button of a stamp on the Passport dashboard
- JSON modals are not showing correctly when there a multiple versions in the DOM

**Expected result**
The app should show a JSON modal including the VC(s) associated with the selected stamp.

**Actual result**
The app opens multiple modals — pinning the JSON modal often GitPOAP stamp.

![Image](https://user-images.githubusercontent.com/59769856/189385672-0ad12543-efcf-433f-898b-ce34ea229417.png)

PR #578, opened 2022-09-12
fix(app): abstracts platformCard to its own component

gdixon:
This PR abstracts `platformCard` to its own component so that the `disclosure` controlling the json modal is unique to the context.

--

Close: #567


farque65 (approved):
🚀 LGTM!

PR #598, opened 2022-09-17
fix(app, iam): modifies lens stamp to verify lens handle

### LINKED ISSUE #595: [BUG] Passport holders should be able to verify the Lens stamp if they have a Lens profile
Scenario
- Passport holder try to verify their Lens stamp

Expected result
- If they have a Lens profile, the stamp should verify succesfully

Actual result
- The stamp only gets verified if they have a Lens profile explicitly set as their default Lens profile

Notes
- It's similar to the difference of having and pointing an ENS domain

farque65:
Closes #595

### DIFF
--- a/app/config/providers.ts
+++ b/app/config/providers.ts
@@ -278,8 +278,8 @@ export const STAMP_PROVIDERS: Readonly<Providers> = {
   ],
   Lens: [
     {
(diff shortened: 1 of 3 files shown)


gdixon (approved):
LGTM!

Issue #580, opened 2022-09-13
[PASSPORT] As a Passport holder, I want to remove a data point from a stamp


erichfi:
GIVEN that I'm a Passport holder and have verified a data point within a stamp
WHEN I toggle off a verified data point
THEN the continue button gets enabled
AND WHEN i hit continue
THEN the data point gets removed from my stamp and Passport, visually and on Gitcoin's Ceramic node

- Designs: https://www.figma.com/file/RIsPkjpIaeEc73SjLiHK2G/Passport?node-id=4287%3A67597
- Context from past implementation: https://github.com/gitcoinco/passport/issues/208

aminah-io:
Removing stamp data points and removing whole stamps is implemented!

aminah-io:
This looks good to test again

PR #599, opened 2022-09-17
feat(app, database-client): adds ability to remove data point and whole stamp

### LINKED ISSUE #579: As a Passport holder, I want to remove a stamp alltogether
GIVEN that I'm a Passport holder and have verified a stamp
WHEN I click the context-menu of the associated stamp on the stamp dashbord
AND when I click remove stamp
THEN the app opens a pop up asking whether I want to really continue removing the stamp
AND WHEN i hit continue
THEN the stamp gets removed from my stamp and Passport, visually and on Gitcoin's Ceramic node

- Designs: https://www.figma.com/file/RIsPkjpIaeEc73SjLiHK2G/Passport?node-id=4287%3A67597
- Context from past implementation: starhttps://github.com/gitcoinco/passport/issues/208

aminah-io:
- [x]  needs an updated `done toast` to reflect user actions
- [x]  needs design for modal popup asking whether user wants to really continue removing the stamp

_to test the stamp action menu, on line 92 of `PlatformCard.tsx` change `selectedProviders[platform.platform].length > 0` to `selectedProviders[platform.platform].length === 0`_

Closes #579 and #580


database-client/src/ceramicClient.ts
```diff
@@ -207,6 +207,28 @@ export class CeramicDatabase implements DataStorageBase {
     }
   }
 
+  // Update stamps by providerIds
+  async updateStamps(providerIds: PROVIDER_ID[]): Promise<void> {
```
gdixon, line 211: Should we rename this `removeStamps()`?

database-client/src/ceramicClient.ts
```diff
@@ -207,6 +207,28 @@ export class CeramicDatabase implements DataStorageBase {
...
+    this.logger.info(`updating stamp(s) on ${this.did}`);
+
+    const passport = await this.store.get("Passport");
+
+    // filter Passport stamp list by the platform's providerIds
+    if (passport && passport.stamps) {
+      const updatedStamps = passport.stamps.filter(stamp => !providerIds.includes(stamp.provider as PROVIDER_ID));
```
gdixon, line 218: As part of the filter, we could also delete the stamps streamId (`await deleteStamp(stamp.streamId)`) so that we're not leaving orphaned streams around?

schultztimothy:
Code wise everything is looking good from what I can see 🚀

One thing I noticed when testing was that I wasn't prompted with a toast to confirm that I wanted to delete my stamp. Was that intended?

aminah-io:
> Code wise everything is looking good from what I can see 🚀
> 
> One thing I noticed when testing was that I wasn't prompted with a toast to confirm that I wanted to delete my stamp. Was that intended?

Hmmm, you should get a remove stamp modal like so:
Screen Shot 2022-09-30 at 2 50 53 PM

And then a done toast after you've hit the `Remove Stamp` button on the modal, like so:
Screen Shot 2022-09-30 at 2 52 22 PM

That's not what you're seeing?

nutrina (changes requested):
Have tested locally, this seems to work as intended.
However, please add integration tests in the database-client

nutrina (approved):
LGTM

PR #614, opened 2022-10-07
feat(app, iam, platforms): adds a generic platform frontend and IAM

farque65:
- This implements the twitter provider

TODO: implement EVM platform with a modified platform front end that can be used for other evm platform


app/pages/DashboardV2.tsx
```diff
@@ -0,0 +1,211 @@
+/* eslint-disable react-hooks/exhaustive-deps */
```
schultztimothy, line 1: Is this still necessary? Looks like a duplicate of Dashboard

app/components/CardListV2.tsx
```diff
@@ -0,0 +1,176 @@
+// --- React Methods
```
schultztimothy, line 1: Should we integrate the new platforms into the existing CardList or modify it so that we can release the updated providers individually?

schultztimothy (commented):
🚀 !

It would be great if we could get the inital platform module fully tested so that as we abstract the rest we have a good base to build off of.

Here is a PR I created to remove the obsolete evm-provider package

app/pages/DashboardV2.tsx
farque65: This was created as a point to build off from the old dashboard. We can remove the old one once the new dashboard is set then keep adding the abstracted provider imports to the new one.

app/components/CardListV2.tsx
farque65: CardListV2 will be the one to work with going forward and we can remove the old cardList

nutrina:
I would propose, to merge PR https://github.com/gitcoinco/passport/pull/643  into this, and then merge this into main.
@schultztimothy I agree we need to fully test, but let's do another GH issue for the test, that we pick up right after the merge.
I think it is important to keep things rolling into main.

PR #640, opened 2022-10-19
chore(app): done/error toast messaging according to designs

### LINKED ISSUE #611: [CHORE] Done/Error toast messaging
As a Passport user, when I select data-points to verify, and I click verify, then I want to see appropriate messaging which reflects the action I just took and the result.

figma:
https://www.figma.com/file/RIsPkjpIaeEc73SjLiHK2G/Passport?node-id=4730%3A62137

aminah-io:
Closes #611


app/components/PlatformCards/GithubPlatform.tsx
```diff
@@ -183,12 +183,19 @@ export default function GithubPlatform(): JSX.Element {
           const initialMinusUpdated = difference(initialVerifiedProviders, updatedVerifiedProviders);
           // Updated providers set minus initial providers set to determine which data points were added
           const updatedMinusInitial = difference(updatedVerifiedProviders, initialVerifiedProviders);
+
+          console.log("imu", initialMinusUpdated);
+          console.log("umi", updatedMinusInitial);
+          console.log("sps", selectedProviders.length);
```
farque65, line 189: These console.logs should be removed

app/components/PlatformCards/DiscordPlatform.tsx
```diff
@@ -233,9 +233,9 @@ export default function DiscordCard(): JSX.Element {
       isClosable: true,
       render: (result: any) => (
         <DoneToastContent
-          title="Verificaton Failed"
+          title="Verification Failed"
           body="Please make sure you fulfill the requirements for this stamp."
-          icon="../../assets/whiteBgShieldExclamation.svg"
+          icon="../../assets/verification-failed.svg"
```
schultztimothy, line 238: The error image wasn't showing for me. I think we need to add a default height to the image in the in [DoneToastContent](https://github.com/gitcoinco/passport/blob/9a386249a39f0101bd670d958d5051f325a129f8/app/components/DoneToastContent.tsx#L34)

with h-6 added to the class I was able to get it to render with ` className="sticky top-0 h-6"`

schultztimothy (commented):
Looking good 🚀 I added one comment for why the error image was showing.

We should keep in mind maybe abstracting those success/failure modals, so we don't have to update each on individually going forward, but I think we can figure that out in the abstraction tickets!

aminah-io:
> Looking good 🚀 I added one comment for why the error image was showing.
> 
> We should keep in mind maybe abstracting those success/failure modals, so we don't have to update each on individually going forward, but I think we can figure that out in the abstraction tickets!

Yes! That is definitely a must

PR #646, opened 2022-10-21
feat(platforms): moves procedures to Platforms/src/Twitter

farque65:


platforms/src/Twitter/procedures/index.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import * as twitterOAuth from "./twitterOauth";
+
+export const router = Router();
+
+export type GenerateTwitterAuthUrlRequestBody = {
+  callback: string;
+};
+
+export type GenerateBrightidBody = {
```
schultztimothy, line 12: Looks like this might have been moved over accidentally

schultztimothy (approved):
lgtm 🚀

platforms/src/Twitter/procedures/index.ts
```diff
@@ -0,0 +1,54 @@
+// ---- Server
+import { Request, Response, Router } from "express";
+
+import * as twitterOAuth from "./twitterOauth";
+import { triggerBrightidSponsorship, verifyBrightidContextId } from "./brightid";
```
aminah-io, line 5: Same question as above--should BrightID be under the Twitter platform?

aminah-io (commented):
Couple of questions about BrightID being under the Twitter platform 👏

platforms/src/Twitter/procedures/brightid.ts
```diff
@@ -0,0 +1,40 @@
+import { verifyContextId, sponsor } from "brightid_sdk";
```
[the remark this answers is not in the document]
farque65: Yes, Bright Id should be moved out of the twitter platform. There needs to be one file defining the router that will be imported by the IAM server. I will move the routing file to a top level file called procedureRouter.

aminah-io (approved):
LGTM 🚀

PR #650, opened 2022-10-25
chore(platforms, app, iam): adds abstractions for Snapshot Platform

### LINKED ISSUE #638: Abstract snapshot platform from Passport app

For further details:

aminah-io:
Adds abstractions for Snapshot Platform
```sh
|- app/components
    |- CardListV2.tsx
|- platforms
    |- index.ts
    |- src/
        |- Snapshot/
            |- Providers/
                |- index.ts
                |- snapshotProposalsProvider.ts
                |- snapshotVotesProvider.ts
            |- App-Bindings.ts
            |- Provider-config.ts
```
----
Closes #638

### DIFF
--- a/app/components/CardListV2.tsx
+++ b/app/components/CardListV2.tsx
@@ -6,7 +6,7 @@ import { PlatformGroupSpec, STAMP_PROVIDERS, UpdatedPlatforms } from "../config/
 
 // Providers
 
-import { TwitterProviderConfig } from "@gitcoin/passport-platforms";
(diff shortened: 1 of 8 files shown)


platforms/.eslintignore
```diff
@@ -7,3 +7,4 @@
 /src/didkit-browser/*
 /__mocks__/**/*
 /__tests__/**/*
+/src/Snapshot/__tests__/**/*
```
nutrina, line 10: I would prefer not to add too many if our own source files to the linter ignore list.

nutrina (changes requested):
Looks good, except I would prefer not to extend the ignore list for the linter.

PR #656, opened 2022-10-26
627 abstract ens platform

schultztimothy:
Fixes: #627


platforms/src/index.ts
```diff
@@ -9,3 +9,5 @@ export {
 } from "./Twitter/providers/TwitterFollowerProvider";
 export { TwitterTweetGT10Provider } from "./Twitter/providers/TwitterTweetsProvider";
 export { TwitterPlatformDetails, TwitterProviderConfig } from "./Twitter/Providers-config";
+
+export * as Ens from "./Ens";
```
farque65, line 13: This export method for ens does make it hard to find the modules in the IAM. This is also leading errors in the CI

iam/src/index.ts
```diff
@@ -48,6 +49,7 @@ import {
   TwitterFollowerGTE1000Provider,
   TwitterFollowerGT5000Provider,
   TwitterTweetGT10Provider,
+  Ens,
```
farque65, line 52: This current import has issues. Also the procedureRouter import should be changed to the following 
import { router as procedureRouter } from "@gitcoin/passport-platforms/dist/commonjs/src/procedure-router";

platforms/src/index.ts
```diff
@@ -0,0 +1,13 @@
...
+export {
+  TwitterFollowerGT100Provider,
+  TwitterFollowerGT500Provider,
+  TwitterFollowerGTE1000Provider,
+  TwitterFollowerGT5000Provider,
+} from "./Twitter/Providers/TwitterFollowerProvider";
+export { TwitterTweetGT10Provider } from "./Twitter/Providers/TwitterTweetsProvider";
+export { TwitterPlatformDetails, TwitterProviderConfig } from "./Twitter/Providers-config";
+
+export * as Ens from "./Ens";
```
farque65, line 13: You should pull in the lateset changes from main. This file was removed. There is another export one directory up

aminah-io (commented):
Just one comment--are you planning to add the App-Bindings file?

schultztimothy:
> Just one comment--are you planning to add the App-Bindings file?

Ah yeah good callout. I asked @farque65 this recently. I'm not seeing `App-Bindings` being used anywhere in the twitter provider. I'm happy to add it in, but was unsure what the purpose of it is

aminah-io:
> > Just one comment--are you planning to add the App-Bindings file?
> 
> Ah yeah good callout. I asked @farque65 this recently. I'm not seeing `App-Bindings` being used anywhere in the twitter provider. I'm happy to add it in, but was unsure what the purpose of it is

Okay cool, it's all good then! 👍👍

farque65:
> Just one comment--are you planning to add the App-Bindings file?

Good call out. The App-Bindings file is only useful if you have front end functions or config objects that you need to import. Not all the providers will need specific functions like the web2 providers to set up a broadcast channel.

farque65 (approved):
LGTM

PR #663, opened 2022-10-27
adds tests for the gitpoap abstraction

aminah-io:
Adds tests for the GitPOAP abstraction from issue #637


aminah-io:
Merged with #637

PR #664, opened 2022-10-27
adds tests for the nft holder abstraction

aminah-io:
Adds tests for the NFT platform abstraction from issue #636


aminah-io:
Merged into #636

PR #679, opened 2022-11-03
feat(app, iam, platforms): adds proof of humanity abstraction

### LINKED ISSUE #630: Abstract Proof of Humanity from Passport app

For further details:

farque65:
Closes #630

### DIFF
--- a/app/next.config.js
+++ b/app/next.config.js
@@ -6,4 +6,8 @@ module.exports = {
     };
   },
   reactStrictMode: true,
+  webpack: function (config, options) {
+    config.experiments = { asyncWebAssembly: true };
+    return config;
+  },
 };
--- a/platforms/package.json
+++ b/platforms/package.json
@@ -18,6 +18,7 @@
       "@ethersproject/base64": "^5.6.1",
       "@ethersproject/providers": "^5.6.2",
       "@gitcoin/passport-types": "^1.0.0",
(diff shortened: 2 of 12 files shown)


platforms/package.json
```diff
@@ -18,6 +18,7 @@
       "@ethersproject/base64": "^5.6.1",
       "@ethersproject/providers": "^5.6.2",
       "@gitcoin/passport-types": "^1.0.0",
+      "@spruceid/didkit-wasm": "^0.3.0-alpha0",
```
nutrina, line 21: Do we need this dependency in platforms?

app/next.config.js
```diff
@@ -6,4 +6,8 @@ module.exports = {
     };
   },
   reactStrictMode: true,
+  webpack: function (config, options) {
```
nutrina, line 9: I assume this is required only for the wasm from spruce?

nutrina (changes requested):
I have some questions ...

platforms/package.json
farque65: Yes, for the getAddress function in poh

nutrina: Ok, that is clear now, after the standup.

nutrina (approved):
LGTM

aminah-io (approved):
LGTM 🚀

PR #683, opened 2022-11-03
feat(platforms, app): refactor logic around initiating issueing a cre…

schultztimothy:
Theres a bug in the callback logic for the callback of accessTokenRequest, this should resolve it and initiate a verification of the Facebook stamp

### DIFF
--- a/app/components/GenericOauthPlatform.tsx
+++ b/app/components/GenericOauthPlatform.tsx
@@ -27,14 +27,14 @@ import { UserContext } from "../context/userContext";
 
 // --- Types
 import { PlatformGroupSpec } from "@gitcoin/passport-platforms/dist/commonjs/src/types";
-import { Platform } from "@gitcoin/passport-platforms/dist/commonjs/src/types";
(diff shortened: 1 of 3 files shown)


app/components/GenericOauthPlatform.tsx
```diff
@@ -155,11 +155,11 @@ export const GenericOauthPlatform = ({
   async function initiateFetchCredential() {
```
nutrina, line 155: What if we move the logic for getting the proof to the providers?
We have already build exception for Twitter and Facebook (and I would hate to add more in the future). I think this component should be as simple as possible and not include logic designed for specific platforms.
I can imagine for example that we request the `accessTokenRequest` for every provider, and the providers can then implement this as required by that specific platform.
And we will only call:
1. accessTokenRequest
2. fetchCredentials
like for the facebook platform.

nutrina (changes requested):
I have some questions, regarding our approach ...

app/components/GenericOauthPlatform.tsx
schultztimothy: I like that idea :) I can start getting that going. So it would basically be a getProof function that exists on the platform that would be called from the front end?

PR #685, opened 2022-11-04
feat(app, iam, platforms): adds poap abstraction

### LINKED ISSUE #631: Abstract POAP platform from Passport app

For further details:

farque65:
Closes #631 

- Moves twitter tests to /platforms/src/Twitter
- Updates Twitter Provider imports in IAM index.ts
- Moves PLATFORM_ID and PROVIDER_ID to platforms/types. Then updates the types used in app


nutrina (approved):
Looks good, tested on local.

PR #686, opened 2022-11-04
634 zksync abstraction

schultztimothy:
fixes: #634

This also includes a change to fix the logic for the facebook provider after a stamp verification is initiated

### DIFF
--- a/app/components/GenericOauthPlatform.tsx
+++ b/app/components/GenericOauthPlatform.tsx
@@ -71,7 +71,7 @@ export const GenericOauthPlatform = ({
   // SelectedProviders will be passed in to the sidebar to be filled there...
   const [verifiedProviders, setVerifiedProviders] = useState<PROVIDER_ID[]>(
     providerIds.filter((providerId) => {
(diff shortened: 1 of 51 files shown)


app/components/GenericOauthPlatform.tsx
```diff
@@ -153,15 +148,14 @@ export const GenericOauthPlatform = ({
   }
 
   async function initiateFetchCredential() {
-    if (accessTokenRequest) {
+    if (platform.getProviderProof) {
```
schultztimothy, line 151: @nutrina Heres an attempt at fixing the comments mentioned here: https://github.com/gitcoinco/passport/pull/683#pullrequestreview-1168179057

This should allow moving the proof logic to platforms for platforms that do not use an Oauth window. 

I tried to extract Oauth window logic to the platforms, but was unable to find an elegant way of doing so. imo it would be cleanest to create a separate `GenericOauthPlatform` component that specifically handles creating the Oauth windows, but very open to suggestions 🚀

nutrina:
@schultztimothy do we still need this PR? Or is this code already on main?

Issue #711, opened 2022-11-09
Abstraction for ClearTextProviders ???


nutrina:

nutrina:
PR is approved, it looks good to me.
@schultztimothy we should however ensure that when pushing this out to prod we also inform to Grants team to test the integration - just in case ...

schultztimothy:
Good call!

PR #714, opened 2022-11-09
feat(platforms): move clear text providers to platforms

schultztimothy:
Fixes: #711


nutrina (commented):
Codewise this looks good to me. 
I am just wondering if we should group the "ClearText" providers together, as this is a special case, not related to the Passport.

schultztimothy:
> Codewise this looks good to me. I am just wondering if we should group the "ClearText" providers together, as this is a special case, not related to the Passport.

this makes sense to me! I moved them over to their own platform

nutrina (approved):
LGTM!

PR #728, opened 2022-11-15
adds functionality to remove passport data points

aminah-io:


platforms/src/GtcStaking/__tests__/communityStaking.test.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
nutrina, line 1: Do we really need this eslint-disable?

nutrina:
I have this error:
image

The problem is that when getting the challenge credential, we must send a type ... however do we need this in the IAM ????

platforms/src/GtcStaking/__tests__/communityStaking.test.ts
aminah-io: I suppose we don't, but I wasn't able to commit my changes until I added it. Is there another flag we can use to bypass eslint for now?

nutrina: We should not bypass eslint, but rather fix the linter errors.

app/components/GenericPlatform.tsx
```diff
@@ -127,39 +127,45 @@ export const GenericPlatform = ({ platFormGroupSpec, platform }: PlatformProps):
...
-        signer as { signMessage: (message: string) => Promise<string> }
-      );
-      // because we provided a types array in the params we expect to receive a
-      // credentials array in the response...
-      const vcs =
-        verified.credentials
-          ?.map((cred: CredentialResponseBody): Stamp | undefined => {
-            if (!cred.error) {
+      // This array will contain all providers that new validated VCs
+      let vcs: Stamp[] = [];
```
schultztimothy, line 131: Nit: might be a smidge cleaner to keep `.map/.filter` https://github.com/gitcoinco/passport/pull/728/files#diff-76a78c07ce4bb2a6bbd9387f6e57db7df025cb43927cb966f5b1eca732c9d097L155 methods as opposed to using let here

schultztimothy (approved):
Works great 🚀 ! I left one little comment but LGTM

app/components/GenericPlatform.tsx
aminah-io: I totally agree! The vcs weren't properly filtering with map and filter though, so we went with the for loop.

PR #730, opened 2022-11-15
feat(app, platforms): enables platform card update pill for Facebook, Twitter and Github stamps

### LINKED ISSUE #589: [BUG] Users should see ‘update’ tags on only Twitter, Github, or Facebook stamps which have they verified before GR15
**Scenario**
- Users should only see ‘update’ tags on Twitter, Github, or Facebook stamps which have they verified before GR15

**Expected result**
Only the Twitter, Github, or Facebook stamp shouldn't show any update tags

**Actual result**
The Google stamp and others from GR14 as well show update tags for Passport holders who verifed these before GR15

farque65:
Closes #589


nutrina (approved):
Code looks good to me.

Issue #621, opened 2022-10-17
Abstract Google platform from Passport app


GTChase:
This story is to break down this parent [ticket](https://github.com/gitcoinco/passport/issues/607) to smaller issues.. The goal is the same, but to abstract just this single platfor from passport.

For further details:
https://github.com/gitcoinco/passport/issues/607

nutrina:
Unable to test this since review and staging environments are not working (IAM service is down).

Troubleshooting he review environment first.

PR #734, opened 2022-11-16
feat(google): this implements google abstraction

nutrina:
This fixes: #621 

In addition it also contains a fix in the database client + related tests


app/components/CardListV2.tsx
```diff
@@ -145,6 +146,14 @@ providers.set("GtcStaking", {
   platFormGroupSpec: GtcStaking.GTCStakingProviderConfig,
 });
 
+providers.set("Google", {
+  platform: new Google.GooglePlatform({
+    clientId: process.env.NEXT_PUBLIC_PASSPORT_GOOGLE_CLIENT_ID,
+    redirectUri: process.env.NEXT_PUBLIC_PASSPORT_LINKEDIN_CALLBACK,
```
schultztimothy, line 152: Looks like this should be google

platforms/src/Google/App-Bindings.ts
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable */
```
schultztimothy, line 1: You should be able to remove this

schultztimothy (approved):
Two tiny comments, but LGTM 🚀

farque65:
I am seeing the following issue when testing locally
Screen Shot 2022-11-17 at 3 54 58 PM

nutrina:
> I am seeing the following issue when testing locally Screen Shot 2022-11-17 at 3 54 58 PM

You need to create a new oauth app, and register the proper redirect URI.
We will need to do this also for the  review, staging and production environments.

app/components/CardListV2.tsx
nutrina: Fixed

PR #742, opened 2022-11-21
fix(iam): adds platforms provider implementation of Brightid

### LINKED ISSUE #740: [Bug] BrightId gives false positives

farque65:
Closes #740


aminah-io (approved):
LGTM 🚀

PR #748, opened 2022-11-23
fix(platforms): resolve path for new docker build

schultztimothy:
Very open to other solutions, this was the best I could think of at the moment for modifying the path to the static html file to work locally as well as within the container


farque65 (approved):
LGTM 🚀

nutrina (approved):

PR #752, opened 2022-11-24
fix(app): adds additional state for data point re-verififications

aminah-io:


nutrina (approved):
Needed to rebase, looks good.

Issue #647, opened 2022-10-24
Unable to verify Gitcoin Grants stamps


GTChase:
**Pre-condition:**
1. donated to GR15 and only GR15

**Steps to reproduce:**
1. claim Contributed in stamp after your GR15 contribution

**Observed Result:**
Stamp is unable to be verified due to a limitation that is in place for contributions needing to be prior to GR15

**Expected Result:**
Stamp should now be able to be verified based on a GR15 contribution

nutrina:
See #680 there is probably a common cause for this.

nutrina:
A PR has been opened: https://github.com/gitcoinco/passport/pull/708

nutrina:
So, apparently the issue is not solved, see: https://discord.com/channels/562828676480237578/1032710228044693565

Status:
   - it is difficult to track what is happening, we do not have sufficient info in the logs for this ... 
   - need to adding more log details then ask the users to test again & check the logs ...

Next steps:
  - return more meaningful and detailed message from IAM (provider) about what verification failed
  - log the verification attempts & the results of these verifications in the Passport app
 
 Expected result:
  - we should see the verification attempts in the datadog logs & be able to determine the cause of the failures

nutrina:
I am blocked because of the issues with the review & staging environments.

PR #753, opened 2022-11-24
fix(#647): adding additional logs to be able to debug issues when ver…

nutrina:
This addresses 647 and other complaints like use not being able to verify google stamps.
This enables checking the errors in the logs.


farque65 (approved):
LGTM

PR #757, opened 2022-11-25
Enhance logs with controller address

nutrina:
This aims to make it possible to debug what is happening when verification fails for some of the providers.


farque65 (approved):
LGTM 🚀

PR #768, opened 2022-12-01
feat(iam, app, platforms): adds humanodeOAuth2Service stamp

sesh92:
Added Humanode private biometric verification stamp which proves user liveness provided via `Humanode OAuth2 Service`


schultztimothy:
👋 @sesh92 thanks for putting this together! I rebased this branch and updated it to be compatible with our updated platform configurations.

Here is the updated PR: https://github.com/gitcoinco/passport/pull/1445 I'm adding a couple comments on that pr if you could please provide more information. Thanks!

PR #783, opened 2022-12-12
Add mainnet info banner for gnosis nft stamps

erichfi:
Adding banners to NFT and Gnosis Safe stamps, highlighting that we only recognize these on Eth mainnet respectively.


nutrina (commented):
Apparently running `yarn lint` on this fails.

PR #794, opened 2022-12-19
feat(iam, app, platforms): add holonym stamp

calebtuttle:
Adds Holonym platform with a provider that checks for uniqueness based on government ID.

For Holonym verification to succeed in Passport, a user must mint a Holo and prove their uniqueness. Users can mint their Holos at https://app.holonym.id/ and can prove uniqueness at https://app.holonym.id/prove/uniqueness.

## Holonym Overview
Holonym is a zero knowledge identity protocol that allows users to prove facts about themselves in a privacy preserving manner. Importantly, Holonym is built around the robust real-world identity systems of government IDs and phone numbers. This architecture allows Holonym to provide strong Sybil resistance guarantees without sacrificing user privacy.

## How does Holonym Improve Gitcoin Passport UX
Holonym was built to make private identity proofs accessible to all internet users, not just web3 power users. Holonym reduces UX friction in the following ways:
Identity verification for Sybil resistance based on familiar methods rather than new methods that involve scanning yourself with an orb, staking tokens, or interacting with complex dapps. Most internet users are already familiar with phone number verification or ID scanning for using apps like airbnb.
Gasless transactions. Users do not need crypto tokens in their wallet in order to produce proofs with Holonym. We simply just require users to have metamask installed to produce signatures.
Quick turnaround times. Users can verify and produce sybil resistance in less than 5 minutes reducing attrition and user drop off.

Tagging the team -- @nanaknihal, @Amosel, @seldamat


Kweiss:
I love this! We are working on our pull request review process and hope to process this request (and others) soon.

thanks for writing this up!

PR #799, opened 2022-12-22
515 mulitple signers

schultztimothy:


aminah-io (commented):
Wow! This was a lot of hard work! Kudos on getting it to this point 🙌 no suggestions here, a lot of improvements have been made and they're awesome! Looking forward to the additional tests and the UI!

iam/src/index.ts
```diff
@@ -414,6 +220,23 @@ app.post("/api/v0.0.0/verify", (req: Request, res: Response): void => {
         // the signer should be the address outlined in the challenge credential - rebuild the id to check for a full match
         const isSigner = challenge.credentialSubject.id === `did:pkh:eip155:1:${address}`;
         const isType = challenge.credentialSubject.provider === `challenge-${payload.type}`;
+
+        // if an additional signer is passed verify that the challenge credential is valid
```
schultztimothy, line 224: @nutrina When you get back from break could you please take a look at this 🙌 ?

app/components/CardList.tsx
```diff
@@ -4,33 +4,9 @@ import React, { useContext, useEffect, useRef, useState } from "react";
 import { PLATFORMS, PlatformSpec } from "../config/platforms";
 import { PlatformGroupSpec, STAMP_PROVIDERS, UpdatedPlatforms } from "../config/providers";
 
-// Providers
```
schultztimothy, line 7: Moved to cermic context so they can be accessed globally within `app`

iam/src/index.ts
```diff
@@ -35,37 +35,8 @@ import {
   verifyCredential,
 } from "@gitcoin/passport-identity/dist/commonjs/src/credentials";
 
-// ---- Identity Provider Management
```
schultztimothy, line 38: updated as single export from platforms

platforms/src/index.ts
```diff
@@ -1,24 +1,225 @@
...
+  Lens,
+  Gitcoin,
+  Github,
+  Facebook,
+  Ens,
+  Twitter,
+};
+
+// Initiate providers - new Providers should be registered in this array...
+export const providers = new Providers([
```
schultztimothy, line 54: Single providers export that is being imported into both `app` and `iam`

PR #844, opened 2023-01-19
Use default CACAO expiry time

oed:
The default CACAO expiry time is 1 week. Currently the passport app has the expiry time set to 1 day. If there is any trouble with the Ceramic anchor service used by the passport Ceramic node, 1 day is a pretty short period to recover from issues. Setting it to the default 1 week gives more room for recovering from potential errors.

### DIFF
--- a/app/context/userContext.tsx
+++ b/app/context/userContext.tsx
@@ -161,7 +161,6 @@ export const UserContextProvider = ({ children }: { children: any }) => {
                 })
               );
               const session = await DIDSession.authorize(authMethod, {
-                expiresInSecs: 24 * 3600,
                 resources: ["ceramic://*"],
               });
               const newSessionStr = session.serialize();


nutrina:
@oed How should I understand this timeout and the CACAO sessions? Is this like a rolling window for the commits being created?
Or does the window to anchor a commit and / or recover from errors get smaller the closer we get to the end of that session expiration time?

oed:
So this value sets the time for how long the given session key is valid. Longer sessions means that on average there is more time for things to get anchored on time.

Of course a commit could be created right before the CACAO is about to expire. For this reason Ceramic considers CACAOs valid for 24 hours after they have been anchored to add some buffer for late anchors.

Let me know if this fully answers your question @nutrina

nutrina:
> So this value sets the time for how long the given session key is valid. Longer sessions means that on average there is more time for things to get anchored on time.
> 
> Of course a commit could be created right before the CACAO is about to expire. For this reason Ceramic considers CACAOs valid for 24 hours after they have been anchored to add some buffer for late anchors.
> 
> Let me know if this fully answers your question @nutrina

Yes, that is clear now. Thx.

Issue #883, opened 2023-01-26
Claim of Stamps in Scorer database via API call


erichfi:
As a Passport holder, I want to claim a VC (JSON) into the Passport Scorer database via an API call

**Acceptance Criteria**

GIVEN a Passport holder is logged in with their Eth address into the Passport app 
WHEN they try to verify a stamp in their Passport
THEN the IAM server tries to verify the conditions for that stamp

GIVEN the IAM server created a valid bearer token to verify that the logged in Passport holder meets the conditions for a specific stamp
WHEN the IAM makes an API call to add a corresponing VC (JSON) into the Passport Scorer database
THEN the API call should return a 201 status code, indicating that the VC (JSON) has been created successfully
AND the system should store the claim VC (JSON) in the Passport Scorer database and associate it with the appropriate owner address and corresponding Passport DID.

--------

GIVEN I am a user that claims a stamp with our IAM server
WHEN the IAM server validates the conditions and creates the Stamp (VerifiedCredential)
THEN it stores the stamp in the DB Cache by posting it to the Scorer API URL
AND then it returns it to the Passport app

GIVEN I am a user that claims a stamp with our IAM server
WHEN the IAM server stores a stamp in the DB Cache
AND a stamp for that address and provider already exists 
THEN the existing stamp will be replaced

TLDR: before the IAM server returns the stamp, it persists it the DB Cache (Scorer API)

The DB model in the DB Scorer fields:
   - id - PK
   - address
   - provider
   - Stamp as JSON Object => this contains the entire VC as JSON

(address + provider) => should be unique

schultztimothy:
As part of https://github.com/gitcoinco/passport-scorer/pull/56 I also added a soft delete functionality after talking with @nutrina . The passport app should now be able to call `DELETE /ceramic-cache/stamp` which will mark the stamp as deleted. Once the `povider` and `address` combination is submitted again, the soft delete column is marked as false

schultztimothy:
https://github.com/gitcoinco/passport/pull/891 Has been marked as a draft

PR #891, opened 2023-01-27
feat(iam): make requests to scorer service to cache stamps after veri…

schultztimothy:
fixes: #883


iam/src/utils/scorer-service.ts
```diff
@@ -0,0 +1,27 @@
...
+      },
+      {
+        headers: {
+          Authorization: `Basic ${process.env.PASSPORT_SCORER_BEARER_TOKEN}`,
+        },
+      }
+    );
+  } catch (e) {
+    // TODO - Log error to DD?
+    console.log({ e });
```
chibie, line 25: could we use `console.error` instead? it would also be great to log the error to datadog!

chibie (changes requested):

Issue #957, opened 2023-02-09
Background Synchronization of Gitcoin Passport Data with Ceramic


erichfi:
In order to ensure that Gitcoin Passport data is available on Ceramic, it is necessary to synchronize it in the background. This synchronization should only occur for Passport holders who are signed in and have an active session.

**Tasks:**
- [ ] Implement a background process that compares the Gitcoin Passport data in the database and Ceramic.
- [ ] If the data in the database is more up-to-date than the Ceramic file, migrate the missing data, including stamps, from the database to the Ceramic file.

**Acceptance Criteria:**
GIVEN a Passport holder is signed in to the app
WHEN we analyze that their Passport in the database is more up-to-date than the Ceramic file
THEN we copy over the missing data from the database to the Ceramic file, ensuring the Ceramic file is up-to-date

PR #965, opened 2023-02-10
feat(app): syncing to ceramic in background

nutrina:
- have added setStamps to CeramicDatabase which should set the content of the Passport on ceramic to contain exactly those stamps

This fixes: #957


app/context/ceramicContext.tsx
```diff
@@ -544,34 +544,43 @@ export const CeramicContextProvider = ({ children }: { children: any }) => {
     database: CeramicDatabase | PassportDatabase,
     skipLoadingState?: boolean
   ): Promise<void> => {
```
nutrina, line 546: Make sure not to load data from ceramic if fetchPassport is not called for the 1st time ....

PR #969, opened 2023-02-13
957 backrgound sync ceramic tests

nutrina:
This adds tests to #957 
This fixes: https://github.com/gitcoinco/passport/issues/978


app/__mocks__/@gitcoin/passport-database-client/index.js
```diff
@@ -1,9 +1,44 @@
 export const createPassportMock = jest.fn();
-export const getPassportMock = jest.fn();
+export const getPassportMock = jest.fn().mockImplementation(() => {
```
schultztimothy, line 2: see if these can be deleted

PR #1057, opened 2023-03-31
feat: implement WIW stamp in gitcoin passport

KevinYum:
Adds a WIW stamp to Gitcoin passport.

This PR implements one basic condition to obtain the WIW stamp based on the number of WIW badges an Ethererum address is eligible for, and the threshold is currently set to >=15.


nutrina:
- [ ]

Issue #1146, opened 2023-04-20
Feature flags should always get the 'on' value


nutrina:
This should make future maintenance much easier, knowing that specifying the `on` value enable a feature, and anything else, will not.
Also it makes deployment of new code safer, as we know, that new (potential experimental features) will not be enabled if one forgets to specify the feature flag.

So here is the convention I propose:

- feature flags are environment variable which will enable a feature at build time (or runtime if possible)
- naming convention:
    - UI apps <framework_refix>_FF_<feature_name>: NEXT_PUBLIC_FF_MULTI_EVM_SIGNER
    - python: FF_<feature_name>
- valid values:
    - the value `on` shoul always enable the feature flag
    - any other value will not enable the feature flag
    - default behaviour (if nothing is specified) => the feature **is not enabled**

PR #1147, opened 2023-04-20
fix(app): feature flags should follow the convention that they alway …

nutrina:
…get an 'on' value.

This closes: https://github.com/gitcoinco/passport/issues/1146


aminah-io (approved):
This looks good, and the proposed convention sounds good as well

PR #1294, opened 2023-05-20
feat: civic pass stamp

dankelleher:
This PR adds [Civic Pass](https://github.com/gitcoinco/passport/pull/civic.com) as a stamp to passport.

Users can get the stamp by obtaining Civic Pass on any of the following chains:

Ethereum Mainnet
Polygon POS
Polygon zkEVM
Arbitrum
XinFin
The supported Civic Passes are:

CAPTCHA
Liveness
Uniqueness
IDV
(see [here](https://docs.civic.com/integration-guides/civic-idv-services/available-networks) for details on each)

Civic Pass stamp verification creates a pop-up to Civic's getpass portal, making it easy for users to get a civic pass directly from the passport UI.

One interface change I had to make - I added the PROVIDER_IDs to the AppContext interface, in order to use them to define an oauth "scope". This could be useful for other providers too I think.


platforms/src/Civic/Providers/civic.ts
```diff
@@ -0,0 +1,136 @@
...
+      };
+    }, {} as Record<string, string>);
+
+    return {
+      valid,
+      error,
+      record: valid
+        ? {
+            address: address,
+            ...recordData,
```
lucianHymer, line 131: The "record" value is hashed in order to create the stamp hash which prevents users from using the same credential (civic pass) with multiple passports. I worry that including this recordData will allow malicious users to e.g. use their civic pass with one passport, then go add it to one more chain, then use their same civic pass (which will have a different hash value) with a second passport. Or, if you guys add support for a new chain and users add this, it would change their stamp hash value. Does that make sense? Do you agree, or am I missing something? I'm thinking we should just use `{address}` for the record value. Thanks!!

dankelleher: Hi @lucianHymer , the passes are associated with a given wallet, so if the user tries to create a second passport they shouldn't be able to use the same passes. Does that make sense? I also might be missing something :)

The reasoning for adding the record data to the credential is that the passes come in various flavours (liveness, uniqueness, full KYC etc), and can be issued on multiple chains. Depending on the types of passes, and the chains they have them on, that might influence the scoring. For example, a uniqueness pass on ethereum would probably be worth more than a liveness pass on mumbai. I didn't want to lose that information.

I have mitigated this to some extent by making the different pass flavours different credentials, but the chain information is still retained inside the credentials themselves. In addition, credentials have metadata such as expiry which may be valuable for scoring.

Issue #1278, opened 2023-05-16
[Spike] Adding OKX Wallet to Blocknative 


Jeremy-Gitcoin:
### User Story:
As an OKX wallet user
I want to be to connect to my OKX wallet
So that I can sign into passport & scorer

### Acceptance Criteria
GIVEN I have an OKX wallet
WHEN I go to Passport or Scorer sites
THEN I am able to select the OKX wallet from available wallets
THEN I am able to sign-in to the sites using my OKX wallet

#### Product & Design Links:
<links>

#### Tech Details:
https://onboard.blocknative.com/docs/wallets/injected#injected-wallets-supported-natively

OKX Docs: https://www.okx.com/web3-docs/extension/introduction

#### Notes/Assumptions:

Let's implement these wallet changes across Passport, Scorer, and Staking.

nutrina:
See this: https://www.okx.com/support/hc/en-us/articles/12833113372941-Access-Blocknative-s-Web3-Onboard-using-the-OKX-Wallet-Web-Extension

lucianHymer:
[Here's the ticket](https://github.com/gitcoinco/passport/issues/1242) about adding a common login flow b/w the apps. This depends on us actually creating a common library, which we still need to make some decisions on. We're leaning towards using it as a subrepo in the existing repos

michaelgreen06:
I have merged [PR 1316](https://github.com/gitcoinco/passport/pull/1316) into main so we can test the behavior of OKX on the whitelisted address of review.passport.gitcoin.co

If the behavior is what we desire then we will need to update the onboard.ts file for scorer.
Staking doesn't use blocknative for login so we will have to investigate if that wallet provider also supports OKX and how it behaves. 

As @lucianHymer mentioned above. Implementing this feature across all our apps would be made easier by having a common login flow.

michaelgreen06:
@erichfi The changes should now be live on review.scorer.gitcoin.co for testing

Please note that staking doesn't have OKX as an option at this time and is using an entirely different wallet provider so it will require some extra work to get it added as an option on staking

Jeremy-Gitcoin:
@michaelgreen06 We need to re-work staking (i'll be talking about this Mon & Tues with the team) so I'm good to ship just for Passport & Scorer!

PR #1316, opened 2023-05-31
1278f okx wallet

michaelgreen06:


michaelgreen06:
addresses #1278 

This may not fully solve to issue though. 

Adding this version to main because OKX added passport.gitcoin.co to a whitelist which blocks its default behavior of taking over for all injected wallets.

Issue #1309, opened 2023-05-26
Integration of Gitcoin Passport Score On-Chain as Part of the Batch Transaction Process


erichfi:
### User Story:
As a Gitcoin Passport holder,
I want to bring both my stamps and the score from my Gitcoin Passport dashboard on-chain,
So that I can verify my digital identity in a more transparent and decentralized manner.

### Acceptance Criteria
GIVEN I am a Gitcoin Passport holder with earned stamps and a score on my Gitcoin Passport dashboard,
WHEN I click on the "Bring my stamps on-chain" button and complete the necessary steps for a batch transaction,
THEN both my stamps and the score from my Gitcoin Passport dashboard are successfully written on the blockchain (Sepolia network) as part of my digital identity, and this information will be reflected in my Gitcoin Passport dashboard.

#### Tech Details:
Our solution to this feature involves several key components and steps that are detailed below:

1. **Data Structure Modification:** We will need to modify the existing data structure used for on-chain transactions. This structure currently holds the stamp data, and we'll extend it to hold the score from the Gitcoin Passport dashboard and the Scorer ID. The new schema will look like this: `{ score: number, scorer_id: bytes32 }`. The 'stamps' will be a hashed representation of the user's verified `PROVIDER_IDS` while 'scorer_id' could be the hashed primary key of the scorer_id in the API.
2. **Encoding EasScore:** We will implement an `EasScore` type, including the EAS encoded representation of the new score schema (`{ score: number, scorer_id: bytes32 }`). This will be incorporated into our data structure as `encodedData`. The EasScore type will look like this:

```
export type EasScore = {
  encodedData: string;
};

export type EasScore = {
  score: EasScore;
  recipient: string;
  expirationTime: any;
  revocable: boolean;
  refUID: string;
  value: any;
  nonce: number;
  fee: any;
};
```

3. **Signature Verification:** We will use an EIP712 signature of the `EasScore` for signature verification. This approach is simpler and more gas efficient than verifying a Verifiable Credential (VC) on-chain.
4. **Smart Contract Modification:** The smart contract that handles the on-chain transactions will need to be updated to handle the new score data. This includes updating the "addPassport" function in the GitcoinAttester contract to accept the new score data as part of the MultiAttestationRequest[] input.
[...]

PR #1338, opened 2023-06-07
1309 score onchain

nutrina:
Fixes: https://github.com/gitcoinco/passport/issues/1309


types/src/index.d.ts
```diff
@@ -182,8 +184,15 @@ export type EasPassport = {
   fee: any;
 };
 
+export type PassportAttestation = {
+  multiAttestationRequest: MultiAttestationRequest[];
+  recipient: string;
```
nutrina, line 189: I think the recipient does not need to be here, since it will always be included in the `MultiAttestationRequest.data`

PR #1353, opened 2023-06-13
feat: adds cyberconnect profile stamp

HaoPeiwen:
3 tiers of cyberprofile stamp


HaoPeiwen:
removed free tier stamp
added cyberconnect org member stamp

platforms/src/CyberProfile/Providers/cyberconnect.ts
```diff
@@ -0,0 +1,127 @@
...
+    let lengthOfPrimaryHandle: number;
+    try {
+      lengthOfPrimaryHandle = await getLengthOfPrimaryHandle(address);
+    } catch (e) {
+      return {
+        valid: false,
+        error: ["CyberProfile provider get user primary handle error"],
+      };
+    }
+    valid = lengthOfPrimaryHandle <= 6 && lengthOfPrimaryHandle > 0;
```
schultztimothy, line 78: Hello @HaoPeiwen thanks for putting this together! I was testing this with a random address that has a handle. Is there a reason we are checking that the handle has a length of less than `6`? For example `hanlapandnamcunguyen` was returned from my test and it seems to be a valid handle

schultztimothy:
Hello @HaoPeiwen thanks a lot for putting this together!

Could you provide a bit more information or links to documentation on the criteria for each stamp? Specifically the `CyberProfilePaid` and `CyberProfilePremium`? It looks like the validity of each stamp is determined by the length of the profile? I was testing with the following profile(`hanlapandnamcunguyen`) and it was unable to qualify for either stamp since the length of the handle was greater than 12.

Thanks!

Issue #1367, opened 2023-06-14
Ensure Inclusion of Credential Issuance Dates in the Passport Attestation Schema


erichfi:
### User Story:
As a developer or contributor working on the On-Chain Passport project,
I want the attestation schema to include the issuance dates for each credential,
So that I can have more granular information about the user's activity and trustworthiness.

### Acceptance Criteria
GIVEN I am a developer or contributor to the On-Chain Passport project,
WHEN I inspect the on-chain Passport attestation,
THEN I should be able to see not only the credentials held by a user but also the dates each of these credentials were issued.

#### Product & Design Links:
https://github.com/gitcoinco/eas-proxy/blob/main/docs/01-on-chain-passport-attestation.md

#### Tech Details:
The enhancement of the schema will require the creation and minting of attestations that include the issuance dates of all the respective credentials.

- [ ] Create a new schema in EAS.
- [ ] Adjust the way we format the attestations. 

#### Open Questions:
- Will the inclusion of issuance dates increase the on-chain transaction costs significantly?

#### Notes/Assumptions:
This feature request is for Option 2 of the Passport Attestation in the On-Chain Passport project. It is assumed that the addition of issuance dates will provide valuable context and history about the passport holder's activity, aiding in more accurate trust assessments.

nutrina:
This is covered in the documentation in this PR: https://github.com/gitcoinco/eas-proxy/pull/20

Docs are here:
- https://github.com/gitcoinco/eas-proxy/blob/on_chain_docs/docs/00-on-chain-data.md
- https://github.com/gitcoinco/eas-proxy/blob/on_chain_docs/docs/01-on-chain-passport-attestation.md

nutrina:
@erichfi given that the issuance date is of importance, shall we also include the expiration date of the individual stamps in the attestation schema?
We do have the option to set an expiration date for the entire attestation, but this would apply for the entire passport.
Instead, I think we should add one in the schema, such that we can track expirations for individual stamps.

nutrina:
Open points for this task:
- [ ] decision if to also include expiration date for individual stamps (similar array as the issuanceDates)
    - [ ] if yes, then the docs also need update
- [ ] review tests + add tests for more scenarios
- [ ] add an equivalent test as `app.post("/api/v0.0.0/eas", (req: Request, res: Response): void => { ...` but for the new stamp
- [ ] call the new endpoint from the Passport app

PR #1415, opened 2023-06-28
feat(iam): add support for passport attestations

nutrina:
This closes: https://github.com/gitcoinco/passport/issues/1367

Updated Schema: https://sepolia.easscan.org/schema/view/0x49d9700408d4ed2c132a3d6410ada3ed794e9d927accfdc02a9ceae75de7af97


iam/src/utils/easPassportSchema.ts
```diff
@@ -0,0 +1,214 @@
...
+  index: number;
+  name: string;
+};
+
+const stampMetadataEndpoint = process.env.PASSPORT_STAMP_METADATA_PATH || "";
+
+export const buildProviderBitMap = async (): Promise<Map<string, PassportAttestationStamp>> => {
+  const stampMetadata: {
+    data: StampMetadata;
+  } = await axios.get(stampMetadataEndpoint);
```
schultztimothy, line 55: Don't use stampMetadata from app. Create a script that generates bitmap format so that it can be versioned

Issue #1402, opened 2023-06-26
Revamp of Twitter Stamp: Updating Verification Credentials


erichfi:
### User Story:
As a Passport Holder,
I want the ability to verify updated credentials about my Twitter activity,
So that my Passport more accurately reflects my activity and longevity on Twitter.

### Acceptance Criteria
GIVEN a Passport Holder with Twitter activity
WHEN I verify my Twitter credentials on my Passport
THEN the following updated verification credentials should be considered:

- Account Age: Account older than 180, 365, 730 days
- Number of days tweeted: More than 30, 60, 120 days

AND

GIVEN a Passport Unique Humanity scorer
WHEN it scores any of these credentials
THEN it assigns a weight of `1.21` to all credentials despite 2 year old accounts and 120 days of tweeting which would get `1.23` respectively.

#### Product & Design Links:
Wording:
 
> **Account Creation:**
> - Created at least 180 days ago
> - Created at least 1 year ago
> - Created at least 2 years ago

> **Consistent Engagement**
> - Tweet on at least 30 distinct days
> - Tweets on at least 60 distinct days
> - Tweets on at least 120 distinct days

#### Tech Details:
The verification system should update its criteria for Twitter based on the new thresholds.

#### Open Questions:
How will these changes affect the existing Twitter verification credentials?

#### Notes/Assumptions:
We are assuming that these new verification credentials will more accurately reflect users' activity and longevity on Twitter.

lucianHymer:
We should add an explicit rate limit check to this stamp's logic--like we did to Github--so that we aren't just showing a generic error to the user when the rate limit is reached

aminah-io:
@erichfi @Jeremy-Gitcoin 
Passport is currently using the Free Twitter v2 API plan, which greatly limits what endpoints we have access to. See the image below for reference:

<img width="1257" alt="Screen Shot 2023-07-05 at 9 35 46 PM" src="https://github.com/gitcoinco/passport/assets/58310685/e100194a-a640-4448-8484-237f2553d7f1">

In order for us to be able to be able to have the stamps that cover a user's days tweeted, we would need to be on the Basic plan at $100, which would give us access to the `timeline` endpoint.

I did some testing in Postman, as well as in the Passport app locally, and I receive a `403` error code when I attempt to access the tweet timeline endpoint:
[...]

PR #1436, opened 2023-07-04
1402 revamp of twitter stamp updating verification credentials

aminah-io:
Fixes: https://github.com/gitcoinco/passport/issues/1402


app/__tests__/components/ProviderCards/TwitterCard.test.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 import { screen, fireEvent } from "@testing-library/react";
-import { TwitterPlatform } from "../../../components/PlatformCards";
+// import { TwitterPlatform } from "../../../components/PlatformCards";
```
schultztimothy, line 3: nit might be worth removing

platforms/src/Twitter/procedures/twitterOauth.ts
```diff
@@ -68,54 +88,125 @@ export const getAuthClient = async (sessionKey: string, code: string, context: T
...
+          context.twitter.valid = true;
+          return {
+            id: userId,
+            numberDaysTweeted: tweetDays.size,
+            valid: true,
+          };
+        }
+        // Respect rate limits by sleeping for a short time after each request
+        await new Promise((resolve) => setTimeout(resolve, 1000));
+      } while (nextToken);
```
schultztimothy, line 177: Seems like an edge case since the user would have to tweet a ton I think 11ish times a day(1400/120=11.6)?  but maybe it would be worth while limiting this to 14 calls to avoid rate limiting?

aminah-io: Yeah, that makes total sense, I'll switch this up!

schultztimothy (commented):
LGTM!

Issue #1277, opened 2023-05-16
Integrate Humanode Stamp to Gitcoin Passport


Jeremy-Gitcoin:
User Story:
As a user of Humanode OAuth
I want to quickly end easily verify my face
So that I can prove I am a unique human and earn more stamps

Acceptance Criteria
GIVEN I have clicked on the Humanode OAuth stamp
GIVEN The stamp will prompt to run through the Oauth flow of doing face scan
WHEN the face scan completes and a valid user Id is returned
THEN I should earn the credential of Humanode Oauth stamp

Verifications / credentials should include:
Proven uniqueness using Humanode Oauth

This should be featured flagged OFF for Production (We'll want to wait until Alex can review and score before turning it on for users)

Product & Design Links:
Tech Details:
Returns a Randomly generated UID the first time it sees each user. This is unique in the system and when the same human tries to Oauth again (different wallet, etc) it will return the same UID it generated for this specific human. They do generate a new set of UIDs when the server resets (1x per year + any time they do disaster recovery) and we cannot tell if the UID is from the current generation of active UIDs or from an older set that should be deprecated. Also if the same user comes back in and Oauths after the UID was reset they will be given a new UID so there's chance a user can have 2 valid stamps (1 UID from before the reset and a new UID after the reset).

Humanode team has created a PR -https://github.com/gitcoinco/passport/pull/768
Open Questions:
Notes/Assumptions:
This passes a risk assessment by Alex and we trust their ability to hold and review Biometric data
This depends on https://github.com/gitcoinco/passport/issues/1124

schultztimothy:
A draft PR has been created that updated the human node integration with out most recent platform configuration. I have a couple questions for the Human node team that were asked in the PR. I also sent an email to them asking for clarification https://github.com/gitcoinco/passport/pull/1445

schultztimothy:
Hi, Tim! Thanks for the review. 

We are still in risk assessment with Jeremy, and after this we would like to update our PR as half a year passed and there were major improvements on all fronts. We also need to introduce timestamps as we discussed with Jeremy. 

Victor

schultztimothy:
I'm still waiting for a response on the PR and/or the email I sent out to the humannode team. Here is a link to the thread in the PR https://github.com/gitcoinco/passport/pull/1445/files/bb9acf4a84b1bbe9329d52a9563827d95c7b2d7c

imammuhammadazis:
[...]

PR #1445, opened 2023-07-07
feat(platforms, app, iam) add human node stamp

schultztimothy:
fixes: #1277


MOZGIII:
Hello from the Humanode Core team! @sesh92 will be helping here from our end.

PR #1446, opened 2023-07-07
fix(app): update walletconnect web3 onbaord due to breaking change

schultztimothy:
https://www.npmjs.com/package/@web3-onboard/walletconnect/v/2.4.0


aminah-io (approved):
Looks good to me. Was only able to test it out with Ledger though, but it connected awesomely. LGTM!

app/utils/onboard.ts
```diff
@@ -12,7 +12,11 @@ const injected = injectedModule();
 
 // web3Onboard modules
 
-const walletConnect = walletConnectModule();
+const walletConnectOptions: WalletConnectOptions = {
+  projectId: process.env.NEXT_PUBLIC_WALLET_CONNECT_PROJECT_ID as string,
```
aminah-io, line 16: I think this might need an `or empty string`
```suggestion
  projectId: process.env.NEXT_PUBLIC_WALLET_CONNECT_PROJECT_ID as string || "",
```

Issue #1481, opened 2023-07-14
Blockchain Network Selection


erichfi:
### User Story:
As a passport holder,
I want to choose a blockchain network to move my passport onto and have the system execute the transaction in my wallet provider,
So that I can make my passport data secure and immutable.

### Acceptance Criteria
GIVEN that the on-chain sidebar is open,
WHEN I view the list of networks,
THEN the "Go" button should only be enabled if my passport hasn't been moved to the selected network yet or if the network and my off-chain passport are out of sync.
AND WHEN I hover over an enabled "Go" button,
THEN a tooltip should appear suggesting that I should switch to that network.
AND WHEN I press the "Go" button,
THEN the transaction flow in my wallet provider should be initiated.

#### Product & Design Links:
<links>

#### Tech Details:

- Implement a list of networks on the sidebar.
- Each item should display the network status.
- Text to switch network (step 1) will trigger the Metamask call to switch. 'Onchain/Go' button (step 2) will initiate the transaction in the user's wallet provider to push the passport onchain to that network.
- If the Passport is already onchain, show 'Complete' instead of 'Switch network or onchain/go' to indicate that the user does not need to sync. (This requires querying and determining if there's a diff with onchain)
- Implement a tooltip that appears when a user hovers over an enabled "Go" button.

We'll need to determine a diff between off chain and onchain so that we can correctly determine the sync

We need to store the configuration for the multiple networks:
- network id
- contract addresses: EAS, verifier (everything we need to call our smart contracts)

#### Notes/Assumptions:

- Assumption is that backend logic for detecting network status and whether a user's passport is in sync with the network is already implemented.
- Tooltip text to be "Switch to [Blockchain Name] Network".

Jeremy-Gitcoin:
@erichfi we'll trigger the switch of networks by the Metamask call (switch network) and then Sync to chain

PR #1522, opened 2023-07-28
1481 follow up

schultztimothy:
fixes: #1481


app/context/userContext.tsx
```diff
@@ -354,17 +354,6 @@ export const UserContextProvider = ({ children }: { children: any }) => {
       setSigner(new Web3Provider(wallet.provider).getSigner());
       // Login to Ceramic
       passportLogin();
-      // attach listener
-      wallet.provider.on("chainChanged", async (chainId: string): Promise<void> => {
```
schultztimothy, line 358: @nutrina I remembered at one point we ran into a jwt issue relating to switching networks on the scorer. I needed to remove this for the on chain functionality and everything seems good to me but just wanted to flag it just in case

PR #1547, opened 2023-08-03
feat(platforms): add IDriss stamp

lennardevertz:
## IDriss Stamp Integration
This pull request aims to integrate the IDriss Membership NFT into Gitcoin Passport in response to user demand. The NFT is acquired via registration for the [IDriss registry](https://www.idriss.xyz/), involving a $10 lifetime protocol fee. Users can currently register Twitter usernames, emails and phone numbers and identifiers must undergo verification as part of the process.
The non-transferrable nature of the NFT emphasizes individual uniqueness. There are currently 13204 NFT holders.

## Verification Results
`verify()` is successful if a user has regsitered with IDriss AND has claimed the membership NFT on [Galxe](https://galxe.com/IDriss/campaign/GCiV3Uwo64).
`verify()` fails when the user is NOT registered with IDriss OR is registered with IDriss but did NOT claim the membership NFT

Open Questions:
- scorer api testing fails on local windows device. How will point attribution be calculated for new stamps?

## Stamp Provider IDrissProvider

- [x] The Stamp provider name is globally unique (and should not have been used previously). This is because the provider name will be included in the record used  to create the Stamp hash:
- [x] The implementation has been unit tested
- [x] Verification payload
    ```json
    valid: true / false,
    record: {
       ... // Unique payload identifying user
    }
    ```
- [x] The attribute `valid` is false if the Stamp verification fails
- [x] The record must contain the following attributes:
    - 1 or more attributes uniquely identifying the user:
        - An ETH address - it should always be lower case (not checksummed)
        - An email (in case of Google for example)
        - The users unique ID (like for Facebook, Twitter, Github)        
- [x] The payload should never contain a field named `pii`
    - This field is reserved for internal use


geoist1:
👍

Issue #1622, opened 2023-08-22
[BUG]: Fix Twitter oauth issue (Passport & Grants Stack)


aminah-io:
The Twitter oauth is missing the scope, which may be preventing users from authenticating with Twitter on both the Passport and Grants Stack apps.

This fix should look into the cause and correct the issue.

PR #1627, opened 2023-08-22
fix(platforms): adds scope to twitter auth flow

aminah-io:
fixes: https://github.com/gitcoinco/passport/issues/1622


nutrina (approved):
Looks good

Issue #1659, opened 2023-08-29
Better error message when bringing stamps on-chain


nutrina:
- [x]  Increase the IAM instance size, remove autoscaling (this also fixes: https://github.com/gitcoinco/passport/issues/1535 )
- [x]  better error message in case getting the score fails.

PR #1660, opened 2023-08-29
fix(iam,infra): add more detailed errors to /eas requests, increase iam task size and disable scaling

nutrina:
Fixes: #1659

### DIFF
--- a/iam/src/utils/scorerService.ts
+++ b/iam/src/utils/scorerService.ts
@@ -1,7 +1,32 @@
 import axios from "axios";
 import { Score } from "./easStampSchema";
+import { AxiosError } from "axios";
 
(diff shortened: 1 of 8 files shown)


iam/src/utils/scorerService.ts
```diff
@@ -14,14 +39,14 @@ export async function fetchPassportScore(address: string): Promise<Score> {
       };
     } = await axios.get(`${scorerApiGetScore}/${address}`, {
       headers: {
-        "X-API-Key": process.env.SCORER_API_KEY,
+        "X-API-Key": "dwefwef",
```
aminah-io, line 42: Should this be pulling from the .env file?

lucianHymer, line 42: Test value?

Issue #1540, opened 2023-08-03
Implement New Credential Issuance Schema in IAM Server


nutrina:
### User Story:
As a developer,
I want to create and deploy a version of the IAM server that is capable of issuing credentials in the new format and adjust the on-chain issuance accordingly,
So that I can test against this production deployment and deploy updates to the Passport application that use the new stamp schema.

### Acceptance Criteria
GIVEN I am a user of the IAM service
WHEN I make a request for a challenge credential
THEN I get back a valid challenge credential using the new signature (as required in the new VC format)

AND

GIVEN I am a user of the IAM service
WHEN I make a request for verifying a user's claim
AND if I present a valid challenge signature
THEN I get back a valid credential in the new schema
AND that credential has the new proof / signature type EthereumEip712Signature2021
AND the VCs hash is created with a separate key (that will be in the old format)

#### Product & Design Links:
[Example implementation for issuing credentials](https://gist.github.com/nutrina/9306e82080d34abc33aa696753c6c4f2)

#### Tech Details:

```json
{
  "@context": [
    "https://www.w3.org/2018/credentials/v1",
    "https://w3id.org/vc/status-list/2021/v1"
  ],
  "type": [
    "VerifiableCredential"
  ],
  "credentialSubject": {
    "id": "did:pkh:eip155:1:0x12FeD9f987bc340c6bE43fD80aD641E8cD740682",
    "provider": "Twitter",
    "metaPointer": "https://github.com/gitcoinco/passport-scorer/blob/main/api/scorer/settings/gitcoin_passport_weights.py",
    "customInfo": {
      "field1": "value"
    },
    "hash": "v0.0.0:AjcRjxx7Hp3PKPSNwPeBJjR21pLyA14CVeQ1XijzxUc=",
    "@context": {
      "customInfo": "https://schema.org/Thing",
      "hash": "https://schema.org/Text",
      "metaPointer": "https://schema.org/URL",
      "provider": "https://schema.org/Text"
    }
  },
  "issuer": "did:ethr:0xd6fc34345bc8c8e5659a35bed9629d5558d48c4e",
  "issuanceDate": "2022-07-19T10:42:24.883Z",
  "proof": {
    "@context": "https://w3id.org/security/suites/eip712sig-2021/v1",
    "type": "EthereumEip712Signature2021",
    "proofPurpose": "assertionMethod",
[...]

PR #1662, opened 2023-08-30
feat(iam): EIP712 signed verifiable credentials

digitalmnt:
fixes: #1540 

To switch over to the new signature type update the env variables in `app` and `iam` to match the example values in this pr


identity/src/credentials.ts
```diff
@@ -126,21 +196,61 @@ export const issueHashedCredential = async (
...
+          hash: `${VERSION}:${hash}`,
         },
-      ],
-      // construct a pkh DID on mainnet (:1) for the given wallet address
-      id: `did:pkh:eip155:1:${address}`,
-      provider: record.type,
-      hash: `${VERSION}:${hash}`,
-    },
-  });
+        // https://www.w3.org/TR/vc-status-list/#statuslist2021entry
```
digitalmnt, line 224: From my understanding this can be added later and should only be included if we would like to a add a revocation status.

identity/src/credentials.ts
```diff
@@ -126,21 +193,61 @@ export const issueHashedCredential = async (
...
+              hash: "https://schema.org/Text",
+              metaPointer: "https://schema.org/URL",
+              provider: "https://schema.org/Text",
+            },
+          ],
+          // construct a pkh DID on mainnet (:1) for the given wallet address
+          id: `did:pkh:eip155:1:${address}`,
+          provider: record.type,
+          metaPointer:
+            "https://github.com/gitcoinco/passport-scorer/blob/main/api/scorer/settings/gitcoin_passport_weights.py",
```
nutrina, line 217: This metapointer should be another record variable, we should not hardcode it.

types/src/index.d.ts
```diff
@@ -59,6 +67,7 @@ export type RequestPayload = {
   jsonRpcSigner?: JsonRpcSigner;
   challenge?: string;
   issuer?: string;
+  signatureType?: string;
```
nutrina, line 70: Let's define a type for this.

identity/src/credentials.ts
```diff
@@ -78,29 +86,87 @@ const _issueCredential = async (
...
+    type: ["VerifiableCredential"],
+    issuer,
+    issuanceDate: new Date().toISOString(),
+    expirationDate,
+    ...fields,
+  };
+
+  const credential = await DIDKit.issueCredential(
+    JSON.stringify(credentialInput, undefined, 2),
+    JSON.stringify(signingDocument, undefined, 2),
```
nutrina, line 112: No need to ident the JSON object here. Removing the `undefined, 2` should work.

Issue #1764, opened 2023-10-10
Move passport resources (IAM service) to the same VPC as the passport-scorer


nutrina:

nutrina:
@larisa17 check the comments in the PR, we still need to add the redis cluster

PR #1853, opened 2023-11-03
1764 Add passport new  infrastructure

larisa17:
This fixes [Move passport resources (IAM service) to the same VPC as the passport-scorer#1764](https://github.com/gitcoinco/passport/issues/1764)


nutrina (changes requested):
We need to add the redis cluster to this infrastructure.
See: https://github.com/gitcoinco/passport/blob/main/infra/production/index.ts#L147-L176

PR #1936, opened 2023-11-21
feat(app,iam): using did session to sign challenges

### LINKED ISSUE #1892: Consolidation of Signatures in Passport App
### User Story:
As a Product Manager at Gitcoin Passport,
I want to consolidate the signature process in the Passport app,
So that users experience fewer signature requests, enhancing overall usability and streamlining interactions.

### Acceptance Criteria
GIVEN a user has an active session and a key for Ceramic is already obtained,
WHEN they perform actions within the Passport app,
THEN the system should utilize the existing key to sign requests without prompting the user for additional signatures, except for on-chain transactions that require explicit user consent.

#### Tech Details:

- Implement a mechanism to check for an existing key for Ceramic during a user's active session.
- Use the existing key to automatically sign requests in the background, reducing the frequency of signature pop-ups.
- Ensure that the user is prompted for a signature only for critical actions, such as on-chain transactions, where explicit consent is necessary.
- Maintain security and user control, especially for on-chain interactions.

#### Open Questions:

- How will this change impact the overall security posture of the Passport app?
- Are there specific scenarios where users should still be prompted for signatures despite having an active session?
- Do we ALWAYS refresh the Ceramic key when the user SIWE? If not, how long before expiry do we want to prompt another signature for the Ceramic key? 

#### Notes/Assumptions:

- This feature assumes that using an existing key for Ceramic is secure and complies with all necessary privacy and security standards.
- The primary objective is to enhance user experience by reducing redundant signature requests, while still maintaining the highest level of security for critical actions.

lucianHymer:
fixes #1892


iam/src/index.ts
```diff
@@ -397,7 +397,7 @@ app.post("/api/v0.0.0/verify", (req: Request, res: Response): void => {
       if (verified && currentIssuer === challenge.issuer) {
         // pull the address and checksum so that its stored in a predictable format
         const address = utils.getAddress(
-          utils.verifyMessage(challenge.credentialSubject.challenge, payload.proofs.signature)
+          await verifyDidChallenge(signedChallenge, challenge.credentialSubject.challenge)
```
lucianHymer, line 400: The grants stack integration is using the old challenge format.
We could ask them to update, it wouldn't change the UX much. But it may not be trivial to change.

We could also see if the signedChallenge was passed in, and if not fall back to the old method?

Should we do something like:
signedChallenge ? verifyDidChallenge(...) : utils.verifyMessage(...)

nutrina: Yes, I think we should do this.
Or use a different endpoint when using the new signing method.
I can look into this

PR #1961, opened 2023-12-01
Fix lerna version in README setup

nasko25:
I was not able to execute
```bash
lerna init
lerna bootstrap
```
with the default version of `lerna` that is installed with `npm install lerna` (v8.0.0), so I updated the command in the README to include the version of lerna specified in `package.json`


gatoypluma:
yea

PR #2070, opened 2024-01-25
Update Holonym stamp

calebtuttle:
Update Holonym stamp to query the government ID Sybil resistance endpoint in the Holonym API instead of directly calling the v1 smart contract. This allows Passport to immediately take advantage of upgrades to the Holonym protocol without having to modify the stamp with each new Holonym version.

The code for this API is open source and can be viewed here: https://github.com/holonym-foundation/holonym-api


nutrina:
Closing, as this has been rebased.
The new PR is https://github.com/gitcoinco/passport/pull/2074

Issue #2252, opened 2024-03-04
Properly catching & alerting for moralis errors


nutrina:
### User Story

Moralis error are caught but they are reported as Redis errors (causing confusing when devs get the PD call).
Fix the error reporting, and create a separate alarm for Moralis issues.

### Acceptance criteria

GIVEN I am a on-duty Passport dev
WHEN an error occurs while getting prices from the Moralis API
THEN I want this to be reported as a "Moralis Error" in PD (and not as a Redis error)

### Tech Details

- what should the monitoring window for the errors be?
- after how many error logs (Moralis read attempts) shall we trigger the PD error? These errors tend to resolve themselves after ~ 5 min

PR #2260, opened 2024-03-05
2252 catch moralis errors

nutrina:
This fixes: https://github.com/gitcoinco/passport/issues/2252

It also adds actions to `okActions`, as indicated in this PD guide: https://support.pagerduty.com/docs/amazon-cloudwatch-integration-guide


Abuchtela (changes requested):
Actions error

PR #2274, opened 2024-03-08
feat(app): when attempting to re-claim a valid credential that is now unclaimable, leave the existing credential in place

### LINKED ISSUE #2272: Prevent Unintentional Deletion of Eth Stamps During Re-verification
### User Story:
As a Gitcoin Passport user,
I want assurance that attempting to re-verify my Eth Stamp will not result in its unintentional deletion due to a temporary lower score,
So that I can confidently attempt re-verification without fear of losing my previously earned achievements.

### Acceptance Criteria
GIVEN a user attempts to re-verify their Eth Stamp in their Gitcoin Passport,
AND their new score is temporarily lower than their previous score,
WHEN the re-verification process is completed,
THEN the Eth Stamp should not be automatically deleted from their passport.

AND

GIVEN a user's Eth Stamp has not expired and they wish to re-verify it,
WHEN the new score is calculated,
THEN the system should compare the new score with the existing one and maintain the stamp unless explicitly deleted or replaced by the user, even if the new score is lower.

#### Tech Details:

- Modify the re-verification logic to ensure that Eth Stamps are not deleted if a user’s new score is lower than their previous score, unless the stamp has expired according to its original expiration date.
- Implement logic to handle cases where the stamp's re-verification does not meet the initial criteria but the stamp is still within its validity period, ensuring it remains intact.

lucianHymer:
fixes #2272


app/components/GenericPlatform.tsx
```diff
@@ -198,16 +198,23 @@ export const GenericPlatform = ({
...
+      const stampPatches: StampPatch[] = platformProviderIds
+        .map((provider: PROVIDER_ID) => {
+          const cred = verifiedCredentials.find((cred: any) => cred.record?.type === provider);
+          if (cred) return { provider, credential: cred.credential };
+          else if (!selectedProviders.includes(provider)) return { provider };
+          else return null;
+        })
+        .filter((patch): patch is StampPatch => Boolean(patch));
 
       await handlePatchStamps(stampPatches);
```
nutrina, line 219: I would add another test to spy and verify the payload of this function: `handlePatchStamps`

nutrina (changes requested):
Only 1 additional test I would add

Issue #2310, opened 2024-03-15
Update GTC Staking Stamp to Include Identity Staking v2 Verification


erichfi:
### User Story:
As a Passport backend developer,
I want to update the GTC Staking Stamp functionality,
So that it captures and verifies stakes from Identity Staking v2 as part of the stamp verification process.

### Acceptance Criteria
GIVEN a user has staked GTC in Identity Staking v2,
WHEN the user's Passport is being verified,
THEN the GTC Staking Stamp should reflect the user's staking activities from Identity Staking v2.

#### Tech Details:

- Integrate Identity Staking v2 smart contract interactions with the Passport's GTC Staking Stamp verification logic.
- Ensure that the Stamp verification process queries both the original and v2 staking contracts.
- Update backend services to handle additional data from Identity Staking v2.
- Modify the Stamp generation logic to incorporate v2 stakes in the computation of staking levels and credentials.
- Retain the ability to distinguish between v1 and v2 stakes if necessary for future functionality or analytics.

#### Open Questions:

- [ ] Will there be any UI changes required on the Passport front-end to accommodate the inclusion of Identity Staking v2 stakes?

#### Notes/Assumptions:

- It is assumed that the Passport system currently has access to the smart contract of Identity Staking v2 for querying stake information.
- We are assuming that the inclusion of v2 stakes will not change the underlying security model of the Stamps.
- We presume there is a clear understanding of how Identity Staking v2 stakes should influence credentials and overall Passport score.

nutrina:
The provider will have to check both staking smart contracts:
- old one (as the current round ends end of May)
- new one (v2)

PR #2346, opened 2024-03-27
2310 id staking v2

larisa17:
This fixes https://github.com/gitcoinco/passport/issues/2310

### DIFF
--- a/platforms/src/GtcStaking/Providers/GtcStaking.ts
+++ b/platforms/src/GtcStaking/Providers/GtcStaking.ts
@@ -131,8 +131,15 @@ export class GtcStakingProvider implements Provider {
         });
 
         // V2
+        let selfStakeV2 = new BigNumber(0);
         resultsV2.forEach((stake: StakeV2) => {
-          stake.staker == stake.stakee ? selfStakesV2.push(stake) : communityStakesV2.push(stake);
+          if (stake.staker == stake.stakee) {
+            if (new Date(stake.unlock_time) > new Date()) {
+              selfStakeV2 = new BigNumber(stake.amount);
+            }
+          } else {
+            communityStakesV2.push(stake);
+          }
         });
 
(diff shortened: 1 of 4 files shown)


platforms/src/GtcStaking/Providers/GtcStaking.ts
```diff
@@ -108,9 +143,17 @@ export class GtcStakingProvider implements Provider {
           }
         }, new BigNumber(0));
 
+        // SelfStake => total
+        const selfStakeV2: BigNumber = selfStakesV2.reduce((totalStake, currentStake) => {
+          if (new Date(currentStake.unlock_time) > new Date()) {
+            return totalStake.plus(new BigNumber(currentStake.amount));
+          }
+        }, new BigNumber(0));
```
lucianHymer, line 151: Does this work if the selfStakesV2 includes an expired stake in the middle of the list? I would think we'd need to do something like
```
const selfStakeV2: BigNumber = selfStakesV2.reduce((totalStake, currentStake) => {
  if (new Date(currentStake.unlock_time) > new Date()) {
    return totalStake.plus(new BigNumber(currentStake.amount));
   }
   return totalStake;
}, new BigNumber(0));
```
Otherwise "totalStake" will be set to undefined when the condition evaluates to false.

The one above should be fixed too (line 140).

platforms/src/GtcStaking/Providers/communityStaking.ts
```diff
@@ -68,6 +70,38 @@ class CommunityStakingBaseProvider extends GtcStakingProvider {
       0
     );
   }
+
+  getCountRelevantStakesV2(communityStakes: StakeV2[], address: string): number {
```
lucianHymer, line 74: This looks great! The V2 endpoint is returning the data in a summary format already, so there's no need to do the totals for each address like we did on V1. So this could be simplified a bit. But it could also be left as-is!

nutrina (approved):
Looks good

PR #2401, opened 2024-04-18
ComposeDB Retry Logic

mzkrasner:
Includes retry logic for ComposeDB single retry (on useEffect lifecycle hook)


app/context/ceramicContext.tsx
```diff
@@ -248,6 +248,35 @@ const startingAllProvidersState: AllProvidersState = Object.values(stampPlatform
...
+export const handleComposeRetry = async (
+  composeStamps: Stamp[],
+  passDbPassport: Passport,
+  ceramicClient?: ComposeDatabase
+): Promise<SecondaryStorageBulkPatchResponse | void> => {
+  if (ceramicClient) {
+    try {
+      // using stamps as the source of truth, filter stamps for where the issuanceDate does not match the composeStamp if there is a composeStamp with the same provider
+      const stampsToRetry = passDbPassport.stamps.filter((stamp: Stamp) => {
+        const existingStamp = composeStamps.find((composeStamp: Stamp) => stamp.provider === composeStamp.provider);
```
stbrody, line 260: this looks like this will only catch stamps that already exist in composedb.  Since stamps don't get updated very often, I imagine the more common form of data discrepancy is going to be a stamp that just doesn't exist at all in composedb, but does in gitcoin's database.  Should we expand this to handle that case as well?

lucianHymer: I thought the same at first! But the question mark in the next line is doing the heavy lifting

lucianHymer: It might make things easier to read if we change
`stamp.credential.issuanceDate !== existingStamp?.credential.issuanceDate`
to
`!existingStamp || stamp.credential.issuanceDate !== existingStamp.credential.issuanceDate`

app/context/ceramicContext.tsx
```diff
@@ -393,6 +434,21 @@ export const CeramicContextProvider = ({ children }: { children: any }) => {
     }
   }, [ceramicClient]);
 
+  useEffect(() => {
+    if (pdbPass && ceramicClient) {
```
lucianHymer, line 438: Since cdbStamps is initialized as an empty array, if pdbPass is loaded before ComposeDatabse.getPassport() is called/completed, then it'll try to write all the stamps to Ceramic.

The easiest thing to do is probably to change it back to similar to how it was in the original PR.

Otherwise, I think it would work if cdbStamps were initialized to undefined and then checked in the useEffect? If you go this route, could you rename cdbStamps and pdbPass to something more descriptive? Maybe initialCeramicStamps and initialPassport?

app/context/ceramicContext.tsx
mzkrasner: Included in commit [d1c7e8b](https://github.com/gitcoinco/passport/pull/2401/commits/d1c7e8b746ed6a0e02edee494ef2067fe8993da0)!

app/context/ceramicContext.tsx
mzkrasner: Thanks @lucianHymer  - I went with the latter (since the original PR was disrupting some of the original tests). Handled in commit [1d17930](https://github.com/gitcoinco/passport/pull/2401/commits/1d1793099b4668189a8ed2d340fe7c458c5275ee)

Issue #2437, opened 2024-04-30
Implement Collapsible Stamp Categories on the Dashboard


erichfi:
### User Story:
As a Passport user,
I want to see my stamps organized into collapsible categories on the dashboard,
So that I can easily manage and view stamps by category according to their relevance and value.

### Acceptance Criteria
GIVEN I am logged into my Passport dashboard,
WHEN I navigate to the "My Stamps" section,
THEN I should see my stamps organized into the following categories: Blockchain & Crypto Networks, Government IDs, Social & Professional Platforms, and Biometric Verification.

AND
WHEN I click on a category header,
THEN the category should expand or collapse, showing or hiding the stamps under that category.

AND
WHEN the categories are displayed,
THEN stamps within each category should be ordered by the number of points they contribute to my total score, from highest to lowest.

AND
WHEN I access the dashboard,
THEN all categories should initially be in an expanded state.

#### Product & Design Links:
Figma Link: https://www.figma.com/file/FOOlLEFaBt87atw6CPkA5P/Passport-App?type=design&node-id=3761%3A36672&mode=design&t=A95YBtu2alZO5RG0-1

#### Tech Details:

- Implement accordion-style collapsible components for each stamp category.
- Utilize existing UI components where possible to maintain a consistent look and feel.
- Ensure that the state (expanded/collapsed) of each category can be controlled by the user.
- Sorting functionality should be dynamic and reflect real-time updates to stamp points.

#### Notes/Assumptions:

- Assumes that all necessary backend APIs for fetching stamps and their categories are in place.
- The user understands that the points displayed are dynamic and may change based on verification status or other factors.
- Stamp to Category Mapping:
  - **Blockchain & Crypto Networks:** Ethereum, NFT, GTC Staking, Idena, Gitcoin, zkSync, Guild, Lens, Snapshot, Gnosis Safe, BrightID, Trusta Labs, ENS
  - **Government IDs:** Coinbase, Holonym
  - **Social & Professional Platforms:** Github, LinkedIn, Google, Discord
  - **Biometric Verification:** Civic

PR #2537, opened 2024-06-03
2437 collapsible stamps 2

larisa17:
This fixes: https://github.com/gitcoinco/passport/issues/2437

### DIFF
--- a/app/components/Category.tsx
+++ b/app/components/Category.tsx
@@ -1,0 +1,128 @@
+import React, { useContext, useEffect, useRef, useState } from "react";
+import { PLATFORMS } from "../config/platforms";
+import { PlatformGroupSpec, STAMP_PROVIDERS } from "../config/providers";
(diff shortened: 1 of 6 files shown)


types/src/index.d.ts
nutrina, line 1: Given that this is used only in app, I would suggest to define this in the app only.

app/components/Category.tsx
```diff
@@ -0,0 +1,172 @@
...
+      const isOpen = openRef.current;
+      if (!isOpen) setPanelMounted(false);
+    }, 150);
+  };
+
+  const handleClick = () => {
+    if (dropDownOpen) handleClose();
+    else handleOpen();
+  };
+
```
nutrina, line 79: Using an `useCallback` ( see https://react.dev/reference/react/useCallback ) on any function defined in the component probably makes sense

Issue #2457, opened 2024-05-14
Implement Robust In-App Notification System for Enhanced User Engagement and Compliance (Frontend)


erichfi:
#### User Story:
As a Passport user,  
I want to receive specific, clear, and actionable in-app notifications for my stamp and Passport activities,  
So that I can maintain an up-to-date and compliant identity verification status efficiently.

#### Acceptance Criteria

**Stamp Expiry Notification**  
GIVEN a user has stamps that are about to expire or have expired,  
WHEN the system detects an expiry,  
THEN the user receives a distinct in-app notification for each expired stamp instructing them to reverify.

**On-chain Passport Expiry Notification**  
GIVEN a user’s on-chain Passport on any blockchain (e.g., Arbitrum) is nearing expiration or has expired,  
WHEN they click on the received expiration notification,  
THEN the system should automatically redirect them to the on-chain sidebar where [the expired status is displayed with a clear 'Update' button](https://www.figma.com/design/FOOlLEFaBt87atw6CPkA5P/Passport-App?node-id=6923%3A2896&t=x4SDsiMrPaOveUvf-1).

**Duplicate Stamp Claim Notification**  
GIVEN a user claims a duplicate stamp for the same account across different Passports,  
WHEN the system identifies the duplicate during verification,  
THEN the user receives a notification detailing which wallet contains the original stamp, with a link to learn more about the deduplication policy, and an advisory that removing the duplicate will not affect their score until the other stamp expires.

**Custom Ecosystem Notification**  
GIVEN the PMs configure custom notifications via Django admin for specific events or promotions,  
WHEN such events occur,  
THEN the user receives a notification with a clear message and a hyperlink directing them to an external site or action.

#### Product & Design Links:
Notification UI Designs: [Figma Design - Passport App Notifications](https://www.figma.com/design/FOOlLEFaBt87atw6CPkA5P/Passport-App?node-id=3761%3A36672&t=x4SDsiMrPaOveUvf-1)

#### Tech Details:

- Implement notifications using a combination of sidebar, toaster, and in-app notification mechanisms to ensure visibility and accessibility.
- Ensure the notification system is integrated with real-time monitoring of stamp and Passport statuses.

#### Notes/Assumptions:

- Notifications should prioritize clarity and simplicity to accommodate ESL users effectively.
- The system should be capable of handling multiple notification types without overwhelming the user, adhering to a well-defined priority and urgency system.
[...]

PR #2603, opened 2024-06-19
2457 notificaitons

digitalmnt:
Fixes: #2457


nutrina (changes requested):
The dropdown menu is not fully visible when only 1 notification exists:

image

app/components/MinimalHeader.tsx
```diff
@@ -16,16 +19,23 @@ const getAssets = () => {
...
+      <OnchainSidebar isOpen={showSidebar} onClose={() => setShowSidebar(false)} />
+      <div className={`flex justify-between h-16 ${className}`}>
+        <div className="flex items-center">
+          <img className="" src={assets.gitcoinLogo} alt="Gitcoin Logo" />
+          <img className="mx-3 md:mx-6" src={assets.logoLine} alt="Logo Line" />
+          <img className="h-8" src={assets.passportLogo} alt="Passport Logo" />
+          <div className="ml-3 text-2xl text-color-1">Passport</div>
+        </div>
+        {verificationState.error !== undefined ||
+          (verificationState.success && <Notifications setShowSidebar={() => setShowSidebar(true)} />)}
```
nutrina, line 36: Also, I would make the notification widget more resilient.
For example if the oneClickVerification fails, I think we should still display the notification bell icon + the notifications we are able to determine in this case.

So I would check the `loading` is false (has finished - maybe we will need more than a boolean here), and regardless of the values of success and error I would display the notifications.

digitalmnt: This is the case currently. If `verificationState.error` is defined the Notifications component will be rendered https://github.com/gitcoinco/passport/blob/eff59719aff497530285ceac4887b74433dbfd5b/app/hooks/useOneClickVerification.tsx#L100. So if an error is thrown during the one click flow the notifications will still be rendered. 

I didn't use the loading key since it starts off as false. There was one case where if a user does not have any stamps, the one click flow exits. I added a success value there so notifications will be rendered.

PR #2624, opened 2024-06-25
2614 scroll endpoints

digitalmnt:
Scroll Service

Fixes: https://github.com/gitcoinco/passport/issues/2640


iam/src/index.ts
```diff
@@ -590,8 +594,16 @@ app.post("/api/v0.0.0/eas/passport", (req: Request, res: Response): void => {
         );
 
         const fee = await getEASFeeAmount(2);
+
         const passportAttestation: PassportAttestation = {
-          multiAttestationRequest,
+          multiAttestationRequest: multiAttestationRequest.map((request) => ({
```
lucianHymer, line 599: Does it make more sense to change the return from formatMultiAttestationRequestWithPassportAndScore? Or is that used elsewhere?

iam/src/index.ts
```diff
@@ -687,3 +706,112 @@ app.use("/procedure", procedureRouter);
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 app.use("/static", express.static(path.join(__dirname, "static")));
+
+const prom = async () => {
```
lucianHymer, line 710: I think this was left in accidentally?

iam/src/index.ts
```diff
@@ -687,3 +706,112 @@ app.use("/procedure", procedureRouter);
...
+// Check Eligibility For Minting Badge
+app.get("/scroll/check", async (req: Request, res: Response): Promise<void> => {
+  const { badge, recipient } = req.query;
+
+  if (!badge || !recipient || typeof recipient !== "string" || typeof badge !== "string") {
+    return void errorRes(res, "Missing badge or recipient parameter", 400);
+  }
+
+  const SCROLL_CHAIN_ID: keyof typeof onchainInfo = "0x82750";
+  const scoreSchema = onchainInfo[SCROLL_CHAIN_ID].easSchemas.score.uid;
```
lucianHymer, line 723: Unused var

iam/src/index.ts
digitalmnt: Good catch. This actually needed removed!

Issue #2663, opened 2024-07-02
Re-verification Failure Should Display "Not Verified Yet" Instead of "Expired"


erichfi:
### User Story:
As a user,
I want the system to show "Not Verified Yet" instead of "Expired" when re-verification fails,
So that I can understand that my stamp is in a non-verified state rather than incorrectly assuming it has expired.

### Acceptance Criteria
GIVEN a user with an expired stamp,
WHEN the user attempts re-verification and fails,
THEN the system should display "Not Verified Yet" instead of "Expired" on the stamp.

#### Tech Details:

- Modify the logic handling re-verification outcomes to distinguish between expired and not yet verified states.
- Update UI components to reflect the "Not Verified Yet" status appropriately.
- Ensure consistency in the messaging across different parts of the application where the stamp status is displayed.

#### Notes/Assumptions:

- Assumes that the backend can distinguish between an expired stamp and a failed re-verification attempt.
- Assumes that this change applies to all types of stamps within the system.

PR #2670, opened 2024-07-03
fix: need to click next twice and verification error

digitalmnt:
fixes: #2663

### DIFF
--- a/app/components/InitiateReverifyStampsButton.tsx
+++ b/app/components/InitiateReverifyStampsButton.tsx
@@ -73,7 +73,7 @@ export const ReverifyStampsModal = ({ isOpen, onClose }: ExpiredStampModalProps)
     return possibleProviders.filter((provider) => expiredProviders.includes(provider)).length > 0;
   });
 
-  const handleClaimStep = async (step: number, error?: PLATFORM_ID | "EVMBulkVerify"): Promise<void> => {
+  const handleClaimStep = async (step: number): Promise<void> => {
     if (status === "in_progress") {
       setCurrentStepInProgress(true);
     } else {
--- a/app/components/InitiateReverifyStampsButton.tsx
+++ b/app/components/InitiateReverifyStampsButton.tsx
@@ -53,12 +53,27 @@ export const ReverifyStampsModal = ({ isOpen, onClose }: ExpiredStampModalProps)
     });
   };
 
+  const errorToast = (platform: string) => {
+    toast({
+      duration: 9000,
+      isClosable: true,
+      render: (result: any) => (
+        <DoneToastContent
+          title="Error"
+          message={`Failed to reverify stamps for ${platform}. Please double check eligibility and try again.`}
+          icon="../assets/verification-failed-bright.svg"
+          result={result}
+        />
+      ),
+    });
+  };
+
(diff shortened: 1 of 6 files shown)


app/components/InitiateReverifyStampsButton.tsx
```diff
@@ -53,12 +53,27 @@ export const ReverifyStampsModal = ({ isOpen, onClose }: ExpiredStampModalProps)
...
+    });
+  };
+
   const expiredPlatforms = Object.keys(STAMP_PROVIDERS).filter((provider) => {
     const possibleProviders = getProviderIdsFromPlatformId(provider as PLATFORM_ID);
     return possibleProviders.filter((provider) => expiredProviders.includes(provider)).length > 0;
   });
 
-  const handleClaimStep = async (step: number, platformId?: PLATFORM_ID | "EVMBulkVerify"): Promise<void> => {
+  const handleClaimStep = async (step: number, error?: PLATFORM_ID | "EVMBulkVerify"): Promise<void> => {
```
lucianHymer, line 76: Was this second param supposed to be dropped? Doesn't look like it's used, and has a weird type.

app/context/stampClaimingContext.tsx
```diff
@@ -69,15 +71,17 @@ export type StampClaimForPlatform = {
 
 export interface StampClaimingContextState {
   claimCredentials: (
-    handleClaimStep: (step: number, platformId?: PLATFORM_ID | "EVMBulkVerify") => Promise<void>,
+    handleClaimStep: (step: number, error?: PLATFORM_ID | "EVMBulkVerify") => Promise<void>,
```
lucianHymer, line 74: Same here

nutrina (changes requested):
This does not seem to work as expected.
After re-verification, the stamp is still highlighted as expired when testing on my local.

app/components/InitiateReverifyStampsButton.tsx
```diff
@@ -100,7 +117,7 @@ export const ReverifyStampsModal = ({ isOpen, onClose }: ExpiredStampModalProps)
       }
     });
 
-    await claimCredentials(handleClaimStep, [evmStampClaim, ...stampClaims]);
+    await claimCredentials(handleClaimStep, indicateError, [evmStampClaim, ...stampClaims]);
```
nutrina, line 121: What is actually required is that after this claiming step, we delete any expired stamps for providers that have not been verified successfully.

For example if we are thinking about the staking stamps, if the user has un-staked, and his stamp expired, after re-verification step (which will now fail), we do not want to show him any more that he has an expired stamp.

nutrina (changes requested):
Please see my comment on the code.

app/components/InitiateReverifyStampsButton.tsx
digitalmnt: Yeah I understand the use case. How are you testing this? If the user no longe qualifies for the stamp `await handlePatchStamps(stampPatches);` will delete the stamp if they do not qualify for it. Since it will be filtered out in.

`const stampPatches: StampPatch[] = selectedProviders.map((provider: PROVIDER_ID) => {
            const cred = verifiedCredentials.find((cred: any) => cred.record?.type === provider);
            if (cred) return { provider, credential: cred.credential as VerifiableCredential };
            else return { provider };
          });`

Issue #2628, opened 2024-06-25
Infra Week: Create Utility Script for Exporting ENV Variables and Update Pulumi Scripts


erichfi:
**User Story:**
As a **developer**, I want to **create a utility script that pulls and exports all relevant ENV variables for any environment to local ENVs and update Pulumi scripts to read these variables**, so that I can **run `pulumi up` from my local environment for review, staging, and production using the correct variables**.

**!! Note from Lucian !!** : I'll update the pulumi scripts as part of https://github.com/gitcoinco/passport/issues/2627, so this ticket can just handle the script for local development. Or really, setting up a .env.local with the right 1password paths and then altering the package.json commands to use the 1password utility, and making sure 1password has good default entries for local dev. See https://developer.1password.com/docs/cli/secrets-environment-variables/

**Acceptance Criteria:**
GIVEN the utility script and updated Pulumi scripts,
WHEN I run the script,
THEN all relevant environment variables should be pulled and exported to my local environment, and the Pulumi scripts should use these variables.

**Tech Details:**
- see https://developer.1password.com/docs/cli/secrets-environment-variables/ for more details
- Develop a script to pull and export environment variables.
- Ensure the script works for review, staging, and production environments.
- Modify the Pulumi script to use external environment variables, including `NEXT_PUBLIC_MAX_LEGACY_ROUND_ID`.
- Document usage instructions.

**Open Questions:**

- [ ]  Should the script handle any environment-specific logic?

- [ ] Are there additional environment variables that need to be considered?

**Notes/Assumptions:**
- Assume developers have local environments set up correctly.
- Assume all required environment variables are available.

lucianHymer:
We probably want to use the methods described here? https://developer.1password.com/docs/cli/secrets-environment-variables

nutrina:
The repos affected:

- passport-data-science
- passport
- passport-scorer
- identity-staking (contracts)
- identity-staking-app
- eas-proxy
- passport-scroll-badge
- passport-scroll-badge-services

lucianHymer:
Note: I'll update the pulumi scripts as part of #2627, so this ticket can just handle the script for local development

digitalmnt:
There are POCs for loading the env variables while running pulumi up locally in the linked PRs. Moving to blocked until 2627 is finished up since there is a decent amount of overlap

digitalmnt:
This is no longer needed

PR #2710, opened 2024-07-18
feat(infra): one password env variables

digitalmnt:
Fixes: [#2628](https://github.com/gitcoinco/passport/issues/2628)


infra/README.md
```diff
@@ -15,18 +15,15 @@ export AWS_PROFILE=user1
...
 ## Configure environment
 
-Before we can run `pulumi up`, there are a few resources that need to be built/pushed to the aws environment, such as the secret manager, Route53 Hosted Zone, and iam-server docker image. Once these are present on aws, we must reference them in our local environment in order for pulumi to be aware of them:
+Install the One Password CLI and follow the isntructions to download the CLI and desktop application. https://developer.1password.com/docs/cli/get-started/#install
+
+This will include authenticating the CLI with your 1Password account.
+
+Once the 1password cli is installed you can run `pulumi up`. All necessary env variables will be fetched from 1password and set in the environment.
+
+There is on exception to this. You should manually update the sha for `DOCKER_GTC_PASSPORT_IAM_IMAGE=op://DevOps/$APP_SECRETS/iam/ECR_BASE_URL/{update_me}` to pint to the image you would like to deploy.
```
lucianHymer, line 24: pint => point

infra/package.json
```diff
@@ -1,7 +1,13 @@
 {
   "name": "infra",
+  "scripts": {
```
lucianHymer, line 3: I think we could just do e.g. "review" as ENV and fill in the e.g. op://DevOps/passport-$ENV-secrets/CI/BLA, since we are keeping this naming convention across all the repos. But also, it's great as-is!

infra/.env.pulumi
```diff
@@ -0,0 +1,28 @@
+# One Password
+# Secrets
+IAM_SERVER_SSM_ARN=op://DevOps/$APP_SECRETS/iam/DATA_SCIENCE_API_URL
```
lucianHymer, line 3: Keeping the secrets/env that are only really relevant to e.g. the lambda or EC2 envs in the relevant app secion, e.g. `/iam/` is a great idea! That way we won't need to delete those when we're coming back through like we talked  about. But I think the ones like "IAM_SERVER_SSM_ARN" or "ROUTE_53_ZONE" that are only used in the CI and not with the e.g. IAM environment probably belong in a `/CI/` section (so that we're not injecting them into the iam environment unnecessarily)

PR #2721, opened 2024-07-24
2684 workflows

larisa17:
This fixes https://github.com/gitcoinco/passport/issues/2684


.github/actions/test/action.yml
nutrina, line 1: If this is not used, it can be deleted.

.github/workflows/ci.yml
```diff
@@ -44,20 +47,3 @@ jobs:
         run: yarn test:ceramic-integration
         env:
           CERAMIC_CLIENT_URL: http://localhost:7007
-
```
nutrina, line 47: I would expect this to not be removed.
If we have a separate action for this, invoke it here

larisa17: So we want to check the provider bitmaps at every PR ?

nutrina (changes requested):
Please see my comments

.github/workflows/ci.yml
nutrina: Yes.

PR #2858, opened 2024-09-12
2787 verify GitHub membership and deduplicate stamp claims   iam updates 1

nutrina:
Fixes: https://github.com/passportxyz/passport/issues/2787


platforms/src/Github/Providers-config.ts
```diff
@@ -43,7 +41,7 @@ providers = [
     threshold: "60",
   }),
   new GithubContributionActivityProvider({
-    threshold: MAX_CONTRIBUTION_DAYS.toString(),
+    threshold: "120",
```
lucianHymer, line 44: I think moving this var to githubClient makes sense, but could we still source it here? Just so we don't forget to update it if we ever change the threshold here.

nutrina: Yes, make sense

nutrina:
Closing this, it has been included in https://github.com/passportxyz/passport/pull/2870

Issue #2844, opened 2024-09-09
Sidebar is not updated automatically when writing passport onchain


Jkd-eth:
When I write my passport onchain a refresh of the passport is required to refresh the content of the sidebar. Otherwise it will still show the old state that will indicate your passport is expired for example.

PR #2880, opened 2024-09-17
2844 sidebar bug

tim-schultz:
Fixes: #2844 

[This](https://github.com/ethers-io/ethers.js/issues/3835) error was being thrown when the transaction was being read. Upgrading ethers appears to resolve it.


OKEAMAH (commented):
Cool

PR #2924, opened 2024-09-27
chore(deployments): update chain

tim-schultz:
Reverts OP sepolia updates made [here](https://github.com/passportxyz/passport/pull/2880/files#)


OKEAMAH (commented):
V

Issue #2958, opened 2024-10-08
Update Badge Designs in Scroll Developer Badge Campaign


erichfi:
### User Story:
As a Scroll Developer,
I want the badge designs to reflect the updated aesthetics as seen in the attached Figma designs,
So that I can provide a more visually appealing and engaging experience for users interacting with the Scroll Developer badge campaign.

### Acceptance Criteria
GIVEN the Scroll Developer badge campaign,
WHEN I view the badge designs in the specified sections of the Passport App flow,
THEN the badge designs should match the updated aesthetics and placements as outlined in the attached Figma file [Figma Link](https://www.figma.com/design/SFzb234NWo6ViBrMD7SWTj/Passport-App?node-id=15413-7705&t=FkzjyShD38bneqW0-1).

#### Product & Design Links:
[Figma Design](https://www.figma.com/design/SFzb234NWo6ViBrMD7SWTj/Passport-App?node-id=15413-7705&t=FkzjyShD38bneqW0-1)

#### Tech Details:

- Ensure compatibility with existing Badge system APIs.
- Update asset files in the design repository.
- Verify badge display across all devices and resolutions.

#### Open Questions:

- [x] Are there any technical limitations in implementing the new designs?

#### Notes/Assumptions:

- Assumes all stakeholders have agreed upon the design changes.
- Assumes resources are available for design integration and testing.

PR #2975, opened 2024-10-15
feat: scroll mobile designs

tim-schultz:
Fixes: #2958


app/components/ScrollCampaign.tsx
```diff
@@ -23,24 +23,37 @@ export interface ProviderWithTitle extends Provider {
   title: string;
 }
 
+interface TopBadges {
```
tim-schultz, line 26: remove

PR #3087, opened 2024-12-02
feat(platforms): adding isDeprecated option to providers, reject verification

### LINKED ISSUE #3034: Deprecation of the OutDid Stamp
### User Story:
As a platform administrator,
I want to deprecate the OutDid stamp,
So that new users cannot access it and we can streamline our badge offerings.

### Acceptance Criteria
GIVEN the OutDid stamp is set for deprecation,
WHEN a new user enters the Passport Dashboard,
THEN they should be unable to see the OutDid stamp.

GIVEN a user currently holds the OutDid stamp,
WHEN the stamp reaches its normal expiry period,
THEN it should automatically expire, and the user should retain the associated points until the expiry date.

#### Product & Design Links:
N/A

#### Tech Details:

- Update the platform logic to prevent new claims of the OutDid stamp.
- Ensure existing stamps continue to function normally until the expiration.
- Maintain historical data on the OutDid stamp for reporting and analytics.

#### Open Questions:

- [ ] How will the change be communicated to existing users and potential new users?

#### Notes/Assumptions:

- Assumes that the OutDid stamp is obsolete and not needed for future platform updates.
- Assumes user points associated with the stamp do not affect other areas of the platform.
- Assumes deprecation aligns with the overall strategy to simplify the score.

lucianHymer:
fixes #3034

### DIFF
--- a/app/components/StampSelector.tsx
+++ b/app/components/StampSelector.tsx
@@ -152,21 +152,26 @@ export function StampSelector({ currentPlatform, currentProviders, verifiedProvi
 
 const useIncludedGroupsAndProviders = (specs: PlatformGroupSpec[]): Record<string, string[]> => {
   const customization = useCustomization();
+  const { stampScores } = useContext(ScorerContext);
 
   const included = useMemo(() => {
     const included: Record<string, string[]> = {};
-    specs.forEach((spec) => {
-      const providers = spec.providers?.map((p) => p.name) || [];
+    specs.forEach(({ platformGroup, providers }) => {
+      included[platformGroup] = providers.reduce((includedProviders, provider) => {
+        const hasCustomWeights = customization.scorer?.weights;
 
-      if (!customization.scorer?.weights) {
-        included[spec.platformGroup] = providers;
-        return;
-      }
+        // Hide if customization uses custom weights and this has a custom weight of 0
+        const checkHiddenByCustomization = () =>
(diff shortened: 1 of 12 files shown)


app/components/StampSelector.tsx
```diff
@@ -152,21 +152,26 @@ export function StampSelector({ currentPlatform, currentProviders, verifiedProvi
...
-        included[spec.platformGroup] = providers;
-        return;
-      }
+        // Hide if customization uses custom weights and this has a custom weight of 0
+        const checkHiddenByCustomization = () =>
+          hasCustomWeights && parseFloat(customization.scorer?.weights?.[provider.name] || "0") <= 0;
+
+        // Hide if deprecated and score is 0
+        const checkHiddenByDeprecation = () =>
+          provider.isDeprecated && parseFloat(stampScores[provider.name] || "0") <= 0;
```
nutrina, line 170: do `checkHiddenByCustomization` and `checkHiddenByDeprecation` need to be functions?
Why not compute the boolean value of the flag and store it in flags like `isHiddenByCustomization` and `isHiddenByDeprecation` ?

lucianHymer: Just to take advantage of short circuiting, especially if we add more conditions

nutrina: The question is what is cheaper?
a. to define a 2 functions on each iteration
b. or use if / else construct instead the short-circuit construct ...

PR #3148, opened 2024-12-17
fix(app): making distinction between technically possible points and points to tell the user about

lucianHymer:
fixes https://github.com/passportxyz/passport/issues/3134


nutrina (approved):
Looks good

PR #3160, opened 2024-12-30
feat(iam): test run for generating nullifiers with mishti

### LINKED ISSUE #3095: Spike: Mishti human keys for credentials
### User Story:
As the Passport team
I want to explore using Mishti network
So that I can replace our dependency on our stored secret key for hashing

Let's spend up to two days on this spike

Let's work with the Mishti network and the Holonym team to figure out how we would replace our stored secret key with decentraliazed human keys. 

Outcome of the Spike:
- Working example (proof of concept) with Mishti on testnet and/or Mainnet with the ability to hash a JSON document (e.g. create and read a hash) which we could then use to begin implementation of upgrading our credential hashes
- Estimates on Stories #3096  & Y and additional requirements as needed
- Work with Daniel to make sure he knows where the existing DevRel needs improvement with Mishti

#### Product & Design Links:
<links>

#### Tech Details:
- what fees do we need to pay? This is important if we want to hash every stamp with Mishti

lucianHymer:
fixes #3095


iam/src/index.ts
```diff
@@ -254,6 +221,7 @@ app.post("/api/v0.0.0/verify", (req: Request, res: Response): void => {
       return void errorRes(res, "Unable to verify payload", 401);
     })
     .catch((error) => {
+      console.log("ERROR", error);
```
lucianHymer, line 224: TODO remove

iam/src/utils/oprf.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  mishtiInitialized = true;
+};
+
+export const recordToNullifier = async ({ record }: { record: ProofRecord }) => {
+  const cleartextNullifier = JSON.stringify(objToSortedArray(record));
+  await initializeMishti();
+
+  const nullifier = await generate_oprf(
+    process.env.MISHTI_CLIENT_PRIVATE_KEY,
```
nutrina, line 37: Is this private key used only to compute the signature required for the actual mishti call?
I was looking at this: https://docs.mishti.network/usage-instructions/making-requests-to-mishti-network

lucianHymer:
This code will be used in https://github.com/passportxyz/passport/pull/3247

PR #3317, opened 2025-02-28
feat(iam): adding new v2 score attestation type

### LINKED ISSUE #3189: Mishti: change EAS & Verax attestation schema
This request affects the following components:
- iam (for creating new payloads)
- eas-proxy (both EAS & Verax)

Depends on: #3187 

### Summary
- we will drop the current EAS schemas in favour of using only 1 attestation that will have the same structure as the score in the new V2 API, with some adjustments:
  - new field will be added `score_id`
- we will need to update the iam package to create the proper payloads
- the functionality should be off by default and enabled by the same feature flag as the one in #3187 
- the contracts in `eas-proxy` need to be adjusted:
  - we want the functions in the https://github.com/passportxyz/eas-proxy/blob/main/contracts/IGitcoinPassportDecoder.sol to remain backwards compatible with the old schemas as much as possible
    - the only exception will be the getPassport function which returns an array of `Credential`) for which we will only populate the provider field from now on
  - we might add other functions we consider important (like `getStamps()`, `getStamps(scorer_id)` , `getFullScore(scorer_id)`... these are just a few ideas at the moment)
- make adjustments for the Verax smart contracts & portal

### Acceptance criteria

GIVEN I have a verified my stamps in the Passport app
WHEN I click the "Bring Passport onchain" button
THEN my current score is written to the blockchain as a EAS attestation
AND I can view the score, and related metadata (providers, weights, deduplication information) in the eAS explorer

GIVEN I have written my score on-chain
WHEN I refresh the passport application
AND I click on the "Mint onchain" button
THEN sidebar opens and the the onchain status displays that my on-chain score is up to date correctly displayed

GIVEN I have written my score on-chain
WHEN I refresh the passport application
AND I claim one more stamp and increase my score
AND I click on the "Mint onchain" button
THEN sidebar opens and the the onchain status is correctly displayed and indicates that my on-chain score is **not** up to date

lucianHymer:
fixes #3189


app/hooks/useSyncToChainButton.tsx
```diff
@@ -62,6 +62,7 @@ export const useSyncToChainButton = ({
           };
 
           const { data }: { data: EasPayload } = await chain.attestationProvider.getMultiAttestationRequest(payload);
+          console.log("attestation data: ", data);
```
lucianHymer, line 65: Remove

PR #3778, opened 2025-10-16
feat(app): add SOG and TCO human points actions

lucianHymer:
Add two new human points action types to the points breakdown system:
- SOG (Seasoned Passport OG)
- TCO (The Chosen One)

These new point types are now displayed in the HUMN points tooltip in the header when users earn them.

Changes:
- Add SOG and TCO to POINTS_BREAKDOWN_KEY type in scorerContext
- Add display logic for SOG and TCO in MinimalHeader points tooltip


lucianHymer:
fixes https://github.com/passportxyz/passport/issues/3776

PR #3824, opened 2025-12-15
feat(app): migrate authentication to SIWE

### LINKED ISSUE #3547: [feat] Update sign-in to passport for smart accounts
## Current status
Signing in and claiming stamps with smart accounts (or universal profiles) is not supported.
This has to do with the fact that we cannot create did sessions in this case.

An alternative solution would be to replace the did session with a JWT token for example that we issue in the iam.

The flow when users connect to the passport app in this case will look like:
1. user clicks the `Sign in with ethereum` -> this will trigger a SIWE process
2. a request will be made to the IAM backend where we validate the SIWE request (for both EOA & smart accounts) and returns a JWT token (tokens should be use a signature with public / private key pair)
3. we store the JWT token in the frontend and use it in any further communication with the servers (scorer & iam)

This will affect the following components:
- Passport App
- Embeds & Passport App 
- Identity package (accept smart accounts signatures / JWT tokens)

### User Story:
As a ...
I want to ...
So that I can ...

### Acceptance Criteria
GIVEN ...
WHEN ...
THEN ...

#### Product & Design Links:
<links>

lucianHymer:
Replace DID/CACAO authentication with SIWE (Sign-In with Ethereum).

Core changes:
- Update datastoreConnectionContext to use SIWE flow
- Remove did-session and @didtools packages
- Add siwe package (^2.1.4)
- Replace createSignedPayload with direct SIWE signing
- Use userAddress instead of DID object
- Support session continuity (existing valid JWTs keep working)

Test cleanup:
- Remove stale @didtools/cacao mocks from 9 test files
- Remove deleted DID import from ceramicContext.test.tsx
- Update test mocks to use userAddress instead of did object
- Delete obsolete helpers.test.ts (createSignedPayload removed)
- Delete obsolete __mocks__ for @didtools/cacao and did-session

This enables smart wallet support (Safe, Argent, etc.) by using EIP-1271 signature verification on the backend.

fixes #3547

### DIFF
--- a/app/components/scroll/ScrollMintedBadge.tsx
+++ b/app/components/scroll/ScrollMintedBadge.tsx
@@ -50,16 +50,16 @@ export const ScrollMintedBadge = ({ badgesFreshlyMinted }: { badgesFreshlyMinted
   const goToLoginStep = useNavigateToRootStep();
   const goToGithubConnectStep = useNavigateToGithubConnectStep();
   const { address } = useAccount();
-  const { dbAccessToken } = useDatastoreConnectionContext();
+  const { dbAccessToken, userAddress } = useDatastoreConnectionContext();
   const { badges, areBadgesLoading, errors, hasAtLeastOneBadge } = useScrollBadge(address);
   const { failure } = useMessage();
 
   useEffect(() => {
-    if (!dbAccessToken) {
+    if (!dbAccessToken || !userAddress) {
       console.log("Access token is not present. Going back to login step!");
       goToLoginStep();
     }
-  }, [dbAccessToken, goToLoginStep]);
+  }, [dbAccessToken, userAddress, goToLoginStep]);
 
   useEffect(() => {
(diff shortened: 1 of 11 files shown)


app/components/scroll/ScrollMintedBadge.tsx
```diff
@@ -50,16 +50,16 @@ export const ScrollMintedBadge = ({ badgesFreshlyMinted }: { badgesFreshlyMinted
...
   const { address } = useAccount();
-  const { did, dbAccessToken } = useDatastoreConnectionContext();
+  const { dbAccessToken } = useDatastoreConnectionContext();
   const { badges, areBadgesLoading, errors, hasAtLeastOneBadge } = useScrollBadge(address);
   const { failure } = useMessage();
 
   useEffect(() => {
-    if (!dbAccessToken || !did) {
-      console.log("Access token or did are not present. Going back to login step!");
+    if (!dbAccessToken) {
```
lucianHymer: userAddress?

PR #3932, opened 2026-07-29
analytics: enable masked session replay on passport at 25%

nanaknihal:
## Why

Replay is currently disabled outright on this app. The rationale in `_app.tsx` was that app.passport.xyz (wallet / KYC / PII) and passport.human.tech (marketing) share one SPA with no clean route boundary, and it deferred masked replay pending host/route separation.

This enables masked replay instead, which makes the route boundary unnecessary. We want funnel diagnosis on the verification flow, and masking gets it without recording content.

## Heads-up: there is currently no masking configured anywhere

Worth stating plainly, because it was the assumption behind enabling this. On project `185989`:

```
session_recording_masking_config : null
session_replay_config            : null
capture_console_log_opt_in       : true
```

And posthog-js masks **input values** by default but **not on-screen text**. So relying on project settings would have recorded rendered PII: names, dates of birth, document numbers, wallet addresses, balances. Hence masking is set in this file, not upstream:

- `maskAllInputs: true`
- `maskTextSelector: "*"` — masks all on-screen text
- `enable_recording_console_log: false` — console on a wallet app can carry addresses, tokens, raw API responses

Masking happens in the browser, so redacted data is never sent to PostHog at all.

## What you get

Layout, click targets, scroll, navigation and timing, with no readable content. That covers the funnel questions (where users stall, misclick, or hit an error) but will not let you read what someone typed.

If you want less masking later, `maskTextSelector` is the knob — but it should not be narrowed without auditing every wallet/KYC screen first.

## Sampling

25% of sessions (~160/day of ~640), tunable via `NEXT_PUBLIC_POSTHOG_REPLAY_RATE`.

Recording starts explicitly rather than via project sampling, so this app owns its own rate: the project-wide rate is being set low to stop waap.xyz (~4,200 sessions/day) from dominating the bill. See holonym-foundation/silk#988. All four `startSessionRecording` override keys are set explicitly; the `true` shorthand covers only `sampling` and `linked_flag`.

## Separately: production looks stale

`disable_session_recording: true` has been on `main` since 2026-06-26 (03f2707), yet app.passport.xyz recorded **5,760 sessions in the last 9 days** (confirmed single-host, not cross-host attribution), plus 814 on passport.human.tech. That suggests production is running a build predating that commit, so the existing privacy control is not live. Worth checking the deploy pipeline independently of this PR.

## Notes

- Not typechecked locally: `node_modules` isn't installed in this worktree. Option names (`maskAllInputs`, `maskTextSelector`, `enable_recording_console_log`) and the `startSessionRecording` signature were verified against the `posthog-js@1.240.4` type declarations, matching the `^1.240.4` range in `app/package.json`.
- Formatting matches `.prettierrc.mjs` (printWidth 120, double quotes, semicolons); prettier itself could not be run without installing deps.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/pages/_app.tsx
+++ b/app/pages/_app.tsx
@@ -27,6 +27,12 @@ import { AutoVerificationProvider } from "../components/AutoVerificationProvider
 
 const GTM_ID = process.env.NEXT_PUBLIC_GOOGLE_TAG_MANAGER_ID || "";
 
+// Fraction of sessions to session-replay. app.passport.xyz runs ~640
+// sessions/day; 25% keeps ~160 recordings/day, enough to diagnose the
+// verification funnel without paying for full coverage. Tunable via
+// NEXT_PUBLIC_POSTHOG_REPLAY_RATE.
+const REPLAY_SAMPLE_RATE = Number(process.env.NEXT_PUBLIC_POSTHOG_REPLAY_RATE ?? "0.25");
+
 // Initialize PostHog (client-side only)
 //
 // Behavioral analytics standard: heatmaps, scroll, and click maps plus autocapture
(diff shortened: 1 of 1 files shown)


nanaknihal:
Two follow-ups from review, both now confirmed with data rather than inferred.

## 1. The `data:` URI image concern does not apply here

I flagged that an uploaded ID document could land in a recording as a `data:` URI. Checked, and there is no such surface in this app:

- No `<input type="file">`, no `getUserMedia`, no dropzone, no `createObjectURL` anywhere under `app/`
- No `<iframe>` in `app/components` or `app/pages`
- "Biometrics" is a **stamp platform** (`app/config/platformMap.ts:219`), not in-app capture. Verification redirects to third-party providers (Civic, Binance, HumanIdKyc, ...)

Document and biometric capture happens off-site on other origins, and `recordCrossOriginIframes` defaults to `false` regardless. So there is no document-preview element to block, and no `blockSelector` addition is needed. What passport renders back is **text** (stamps, scores, addresses), which `maskTextSelector: "*"` already covers.

This PR is complete as-is on masking.

## 2. Production is confirmed stale, not just suspected

Over the last 7 days, `app.passport.xyz` produced **57,827 events** carrying:

```
site = null, product = null, surface_type = null
```

The 2026-06-26 commit (03f2707) registers `site: "passport"`, `product: "passport"`, `surface_type: "app"`. **Zero** events carry those values. Production definitively predates that commit.

So deploying this PR also ships 5+ weeks of accumulated changes. That is a deploy-planning question independent of this diff, but it should not be a surprise on merge day.

## 3. Incidental: the shared-SPA premise looks outdated

`passport.human.tech` events are tagged `site=human_tech, product=human_tech, surface_type=marketing` — that is the marketing site's instrumentation, not this SPA's. The marketing surface appears to have already moved, which means the original "one SPA, no clean route boundary" rationale for disabling replay no longer describes reality. The surfaces are already separated.

Not a blocker, but the comment this PR rewrites was reasoning from a topology that has since changed.

Minor taxonomy nit while here: 11,036 `app.passport.xyz` events carry `site = "passport.human.tech"` (a hostname where the other surfaces use a short slug). Worth normalizing separately.

PR #3937, opened 2026-08-25
fix(onchain): surface RPC failures instead of silently returning empty

0xblckmrq:
## Summary

`getPassport()` had a silent `catch { return [] }` and `getAttestationData()` caught all errors, logged, and returned `undefined`. Callers converted `undefined` → `{ score: 0, providers: [] }`, which is indistinguishable from a wallet that hasn't minted. React Query never saw `isError`, never retried, and the UI incorrectly showed a "not minted" state for users who have minted.

**Datadog evidence** (7d): ~4k events across ~297 sessions — `AbortError`, `TypeError: Failed to fetch`, `ContractFunctionExecutionError` all hitting the same swallowed catch.

## Changes

**`app/utils/onChainStamps.ts`**
- Removed `try/catch { return [] }` from `getPassport` — errors propagate up
- Removed `try/catch` + inline Datadog logging from `getAttestationData`; return type changed to `Promise<AttestationData>` (non-nullable). Logging moves to the hook so it fires **once after retries exhaust**, not on every attempt

**`app/hooks/useOnChainData.tsx`**
- Added `retry: 3` + `placeholderData: keepPreviousData` per query — transient failures retry 3× and previous data is shown during re-fetch
- Added `isError: boolean` to `OnChainData` interface and threaded through the return value
- Updated Datadog logging `useEffect` with `{ operation, chainId, address }` metadata

**`app/components/DashboardValidStampsPanel.tsx`**
- When `isError`: show "Couldn't load on-chain data — Try again" (calls `refresh()`) instead of stamp list + Mint button, so users aren't shown a false "not minted" state

**`app/__tests__/hooks/useOnChainData.test.tsx`**
- `getAttestationData` throws → `isError` is `true`
- `getPassport` throws (propagated through) → `isError` is `true`
- Legitimate empty data (score 0, no providers) → `isError` is `false` (regression guard)

Fixes holonym-foundation/internal-docs#2543

cc @calebtuttle

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/app/components/DashboardValidStampsPanel.tsx
+++ b/app/components/DashboardValidStampsPanel.tsx
@@ -43,7 +43,7 @@ const StampsList = ({ className, onChainPlatformIds }: StampsListProps) => {
 
 export const DashboardValidStampsPanel = ({ className }: { className: string }) => {
   const { verifiedPlatforms, allProvidersState } = useContext(CeramicContext);
-  const { activeChainProviders } = useOnChainData();
+  const { activeChainProviders, isError, refresh } = useOnChainData();
 
   const hasOnchainProviders = useCallback(
     (platformId: PLATFORM_ID) => {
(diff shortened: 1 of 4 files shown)


app/hooks/useOnChainData.tsx
```diff
@@ -139,6 +136,8 @@ const useOnChainDataQuery = (address?: string) => {
             customScorerId,
             publicClient: publicClient!,
           }),
+        retry: 3,
+        placeholderData: keepPreviousData,
```
calebtuttle, line 140: `keepPreviousData` also applies when the query key changes. Because `address` and `customScorerId` are in this key, switching wallets or scorer configurations can temporarily expose the previous wallet’s on-chain stamps as the new wallet’s data while the new queries retry (and a disabled query after disconnect may retain that placeholder). Same-key refetches already keep cached data, so can we remove this option or gate the placeholder on the previous and current address/scorer being identical?

app/components/DashboardValidStampsPanel.tsx
```diff
@@ -69,11 +69,25 @@ export const DashboardValidStampsPanel = ({ className }: { className: string })
     >
       <div className="my-2">Valid Stamps</div>
       <div className="h-[2px] w-full bg-gradient-to-r from-background via-foreground-2 to-background" />
-      <StampsList className="m-6" onChainPlatformIds={onChainPlatformIds} />
-      <InitiateOnChainButton className="mb-2" />
-      <span className={`mb-2 text-sm ${anyOnchain ? "visible" : "invisible"}`}>
-        <OnchainMarker /> = Onchain
-      </span>
+      {isError ? (
```
calebtuttle, line 72: `isError` aggregates failures from every enabled-chain query, but this panel otherwise represents only `activeChainProviders`. In the default multi-chain configuration, an outage on any one RPC (for example Linea while the user is connected to Base) will therefore hide valid active-chain stamps and the mint controls. Can we preserve error state per chain and have this branch check only the connected chain’s error?

